PR #2, opened 2018-06-26
Add ServiceBus inteface definition

elprans:


pragnagopa:
Can you add the test code as well? I need to look at how this is being used

elprans:
Test code is in the worker repo.  These are abstract definitions.

azure/functions/_servicebus.py
```diff
@@ -0,0 +1,90 @@
+
+import abc
+import datetime
+import typing
+
+
+class ServiceBusMessage(abc.ABC):
```
pragnagopa: Azure Functions supports [ServiceBusQueues](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-queues-topics-subscriptions#queues) and [ServiceBusTopics](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-queues-topics-subscriptions#topics-and-subscriptions). Add ServiceBusTopic

pragnagopa: Rename to ServiceBusQueueMessage

elprans: Do topics and queues actually receive different data?  The binding documentation mentions only that the parameter type may receive instances of [Message](https://docs.microsoft.com/dotnet/api/microsoft.azure.servicebus.message)

pragnagopa: @asavaritayal  - Can you look into this and confirm if we need different types for queue message and topic?

asavaritayal: I don't think we need different types. `function.json` should take care of the exact source.

PR #4, opened 2019-01-09
Make azure namespace compatible with the style of azure-nspkg

elprans:
azure-nspkg uses pkg_resources.declare_namespace which is incompatible
with PEP 420-style namespaces, which may lead to issues if other Azure
packages are installed in a different location.

See Azure/azure-functions-python-worker/268 for details


asavaritayal:
@ankitkumarr @elprans is this ready to merge?

elprans:
This has already been merged actually.

PR #16, opened 2019-06-17
Move rich binding implementations from the worker

elprans:
Per Azure/azure-functions-python-worker#409, the rich binding
implementations are now in the library.

### DIFF
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -6,8 +6,23 @@ from ._http import HttpRequest  # NoQA
 from ._http import HttpResponse  # NoQA
 from ._queue import QueueMessage  # NoQA
 from ._servicebus import ServiceBusMessage  # NoQA
+from .meta import get_binding_registry  # NoQA
+
+# Import binding implementations to register them
+from . import blob  # NoQA
+from . import cosmosdb  # NoQA
+from . import eventgrid  # NoQA
+from . import eventhub  # NoQA
+from . import http  # NoQA
+from . import queue  # NoQA
+from . import servicebus  # NoQA
+from . import timer  # NoQA
+
 
 __all__ = (
+    # Functions
+    'get_binding_registry',
+
     # Generics.
     'Context',
     'Out',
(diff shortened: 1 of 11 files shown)


azure/functions/_thirdparty/typing_inspect.py
```diff
@@ -0,0 +1,375 @@
...
+def _gorg(cls):
+    """This function exists for compatibility with old typing versions."""
+    assert isinstance(cls, GenericMeta)
+    if hasattr(cls, '_gorg'):
+        return cls._gorg
+    while cls.__origin__ is not None:
+        cls = cls.__origin__
+    return cls
+
+
```
pragnagopa: This looks like a general utility. Shoudn't it be part of the python worker?

azure/functions/__init__.py
```diff
@@ -6,8 +6,23 @@
...
+from . import eventhub  # NoQA
+from . import http  # NoQA
+from . import queue  # NoQA
+from . import servicebus  # NoQA
+from . import timer  # NoQA
+
 
 __all__ = (
+    # Functions
+    'get_registry',
```
pragnagopa: Rename to be more explicit. TriggerMetadataRegistry?

azure/functions/_thirdparty/typing_inspect.py
elprans: It already is, but I'd like to avoid adding a dependency on the worker.

elprans:
>  Will you be porting unit tests as well?

We need to figure out the dependency story first.  Tests depend on the worker.

pragnagopa:
> We need to figure out the dependency story first. Tests depend on the worker.

E2E tests depend on worker. Unit tests should be moved here. Separate PR is ok.

maiqbal11:
@elprans, we'll need to do a release with this change in order to run the tests for https://github.com/Azure/azure-functions-python-worker/pull/409 in the CI?

elprans:
@maiqbal11 Yes, we need to merge this and do a release.

maiqbal11:
> @maiqbal11 Yes, we need to merge this and do a release.

Are we blocked on performing the merge for this? Would be good to get this in so we can start seeing CI builds for the generic bindings PR.

elprans:
@maiqbal11 Published azure-functions-python-library-1.0.0b5

PR #21, opened 2019-07-11
Enable CI builds on Azure pipelines

maiqbal11:
Enabling CI builds for this repo with existing tests.


maiqbal11:
@Hazhzeng, verified that this works on my own CI. Initial commit to enable CI will not have a run FYI.

PR #22, opened 2019-07-12
Additional parsing for fractional date times

maiqbal11:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/460

### DIFF
--- a/azure/functions/meta.py
+++ b/azure/functions/meta.py
@@ -184,20 +184,23 @@ class _BaseConverter(metaclass=_ConverterMeta, binding=None):
         # UTC ISO 8601 assumed
         formats = [
             '%Y-%m-%dT%H:%M:%S+00:00',
+            '%Y-%m-%dT%H:%M:%S-00:00',
             '%Y-%m-%dT%H:%M:%S.%f+00:00',
+            '%Y-%m-%dT%H:%M:%S.%f-00:00',
             '%Y-%m-%dT%H:%M:%SZ',
             '%Y-%m-%dT%H:%M:%S.%fZ',
         ]
         dt = None
 
-        too_fractional = re.match(r'.*\.\d{6}(\d+)Z', datetime_str)
+        too_fractional = re.match(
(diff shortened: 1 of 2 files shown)


craiglytle:
This seems to be a show stopper for ANY Python-based Functions.  Is there a work around?

maiqbal11:
@craiglytle, could you elaborate more on how it is affecting all Python-based functions for you? Currently, we are only aware of this being an issue when running from Portal.

craiglytle:
Thanks for the info.  I dig some deeper digging and YES it sort of works when triggered outside of the "Functions Portal".  
But the Functions Monitor page doesn't show any results which makes it hard to see the positive results.  I only see the success by carefully read the logs.  (Perhaps a different issue)
Here's some details:

I'm running the basic 'quick start' for Functions using Python and a Queue trigger.  No other code.

Running Locally: 
- Trigger from Azure Queue Portal ('+ Add Message' button) - **SUCCEEDS**
- Trigger from my app (which adds a message to the same queue) - **SUCCEEDS**

Running on Azure: (After 'func azure functionapp publish <APP_NAME>')
- Trigger from Azure Functions Portal ('> Run' button) - **FAILS** with time stamp error
(e.g. Exception: ValueError: time data '9999-12-31T23:59:59.9999999+00:00' does not match format '%Y-%m-%dT%H:%M:%S.%fZ')
- Trigger from Azure Queue Portal ('+ Add Message' button) - **SUCCEEDS**
... BUT, it doesn't show up as a successful run in the Azure Functions Portal's "Monitor" panel.  I only see it's successful by carefully reading the logs.   
- Trigger from my app that adds message to the queue - **SUCCEEDS**
... BUT, same as above 

So, I concur that this is only an issue when running from the Functions Portal.  
I'll carry on with development.  Thanks.

wkeithvan:
@Hazhzeng Any idea when this will be patched? [Issue 460](https://github.com/Azure/azure-functions-python-worker/issues/460) is still occurring...

wkeithvan:
@maiqbal11 @Hazhzeng Is the fix live on Azure right now? I am still running into the issues in issue [#460](https://github.com/Azure/azure-functions-python-worker/issues/460)

PR #37, opened 2020-01-01
Fix local time parsing

Hazhzeng:
### Related
Resolves https://github.com/Azure/azure-functions-python-worker/issues/591

### Issue
The Azure Media Service for EventGrid trigger metadata's event_time is a local time, which is not in ISO 8601 format. This PR fixes the local time parsing issue.

### DIFF
--- a/azure/functions/meta.py
+++ b/azure/functions/meta.py
@@ -224,13 +224,11 @@ class _BaseConverter(metaclass=_ConverterMeta, binding=None):
             '%Y-%m-%dT%H:%M:%S.%fZ',
         ]
 
-        is_succeeded, dt, _ = try_parse_datetime_with_formats(
+        dt, _, excpt = try_parse_datetime_with_formats(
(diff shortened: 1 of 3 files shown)


azure/functions/meta.py
```diff
@@ -181,37 +181,79 @@ def _parse_timedelta_metadata(
...
+
+        # Local time assumed
+        # 2018-08-07T23:17:57.461050
+        local_formats = [
+            '%Y-%m-%dT%H:%M:%S.%f',
+            '%Y-%m-%dT%H:%M:%S',
+        ]
 
-        for fmt in formats:
+        dt = None
```
maiqbal11: nit: I'd refactor the `_parse_datetime_local` and `_parse_datetime_utc` functions to use a third function that encapsulates the `striptime` logic since that is the same across both. The only variation is the formats which could be passed in as separated arguments.

PR #40, opened 2020-01-27
Allow null value when blob doesnt exist

kolafsson:
Returns None when a customer tries to access blob that doesn't exist

### DIFF
--- a/azure/functions/blob.py
+++ b/azure/functions/blob.py
@@ -75,6 +75,9 @@ class BlobConverter(meta.InConverter,
 
     @classmethod
     def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
+        if data is None or data.type is None:
+            return None
+            
         data_type = data.type
 
         if data_type == 'string':


anthonychu:
@Hazhzeng @anirudhgarg Could you please take a look at this PR? Thanks!

azure/functions/blob.py
```diff
@@ -75,6 +75,9 @@ def encode(cls, obj: typing.Any, *,
 
     @classmethod
     def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
+        if data is None or data.type is None:
+            return None
+
```
Hazhzeng: Hi @kolafsson, thanks for contributing to azure-functions-python-library

Please add a few unit tests regarding to this scenario.
Unit test examples: https://github.com/Azure/azure-functions-python-library/pull/28/files#diff-c27162ae93fd2aa43a783ef13d1749f4R9-R12

Thanks.

Hazhzeng:
Moving to PR #67

PR #42, opened 2020-02-02
Restore IoTHub metadata in EventHub

Hazhzeng:
Resolves https://github.com/Azure/azure-functions-python-worker/issues/418

1. Fix a parsing issue where the iothub metadata is not handled properly. The iothub_metadata does not exist in trigger_metadata, instead, it is placed in **SystemProperties (cardinality = one)** and **SystemPropertiesArray (cardinality = many)**
2. Fix an issue where the IoTHub event messages are treated as single message. Since the message format is `json`, thus, it is handled with `decode_single_event`. This PR will make sure the cardinality is handled correctly.

### DIFF
--- a/azure/functions/eventhub.py
+++ b/azure/functions/eventhub.py
@@ -199,7 +199,7 @@ class EventHubTriggerConverter(EventHubConverter,
 
     @classmethod
     def _decode_iothub_metadata(
-        cls, trigger_metadata) -> typing.Dict[str, str]:
+            cls, trigger_metadata) -> typing.Dict[str, str]:
         # Try extracting iothub_metadata from trigger_metadata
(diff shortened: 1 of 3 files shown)


azure/functions/eventhub.py
```diff
@@ -105,12 +105,10 @@ def decode(
                       typing.List[_eventhub.EventHubEvent]]:
         data_type = data.type
 
-        if (data_type == 'string' or data_type == 'bytes'
-                or data_type == 'json'):
+        if cls._is_cardinary_one(trigger_metadata):
```
AnatoliB: Shouldn't this be "cardinality"? Or is "cardinary" a term I'm not aware of?

Hazhzeng:
Resolves: https://github.com/Azure/azure-sdk-for-python/issues/8533

PR #48, opened 2020-02-21
Add orchestration trigger for durable functions

Hazhzeng:
### Background
To support durable function running in Python Worker, we should create a rich binding for OrchestrationTrigger

### Features
1. Added an OrchestrationContext type for accepting orchestrate context
2. Allow user code to use `orchestration_context.body` to access the orchestration data
3. Set `has_implicit_output=True` for OrchestrationTrigger to enable context return even when `$return` is not defined in function.json

### Usage
```javascript
// function.json
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "name": "orctx",  //context is a reserved keyword in azure functions Python, we use context to pass the function_name and invocation_id to our customers
      "type": "orchestrationTrigger",
      "direction": "in"
      // remove datatype here
    }
  ],
  "disabled": false
}
```

```python
# main.py
import azure.functions as func

def main(orctx: func.OrchestrationContext):
# or
# def main(orctx):
    orchestrate = df.Orchestrator.create(generator_function)
    result = orchestrate(orctx.body)  # need to add .body
    return result
```

### DIFF
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -7,6 +7,7 @@ from ._http import HttpResponse  # NoQA
 from ._http_wsgi import WsgiMiddleware # NoQA
 from ._queue import QueueMessage  # NoQA
 from ._servicebus import ServiceBusMessage  # NoQA
+from ._durable_functions import OrchestrationContext  # NoQA
 from .meta import get_binding_registry  # NoQA
 
 # Import binding implementations to register them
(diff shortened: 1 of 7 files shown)


Hazhzeng:
Hey @scgbear,

This PR works fine with fan-in-fan-out and function-chain.
However, when it uses call_activity_with_retry("ParrotValue", retry_options, i) it always passes a None value to the activity function.

@asedighi and I found out the variable name `input` does not match to its attribute name `self.input_` in the following line https://github.com/Azure/azure-functions-durable-python/blob/7df1ef752b591dbd357f39d813515394796c3104/azure/durable_functions/models/actions/CallActivityWithRetryAction.py#L33

PR containing the fix is here https://github.com/Azure/azure-functions-durable-python/pull/57/files#diff-f44dfb86028dc2811e9365c215fe8e2cR33

PR #53, opened 2020-04-02
Add activity trigger encode/decode logic for durable function

Hazhzeng:
### Dependency
https://github.com/Azure/azure-functions-python-worker/pull/643

### Background
Previously, the Activity Trigger in Python Durable Functions only supports string as return value. Now, we should support JSON serializable Python primitive types.

✔string, int, float, bool, list of string/int/float/bool, dictionary of string/int/float/bool, None
❌set, frozenset, bytes, bytearray, custom type, custom dataclass

### DIFF
--- a/azure/functions/durable_functions.py
+++ b/azure/functions/durable_functions.py
@@ -72,7 +72,7 @@ class ActivityTriggerConverter(meta.InConverter,
                     f'valid json serializable ({data.value})')
         else:
             raise NotImplementedError(
-                f'unsupported event grid payload type: {data_type}')
+                f'unsupported activity trigger payload type: {data_type}')
 
         return result
 
--- a/azure/functions/durable_functions.py
+++ b/azure/functions/durable_functions.py
@@ -1,4 +1,6 @@
-from typing import Any
+import typing
+import json
+
 from azure.functions import _durable_functions
 
 from . import meta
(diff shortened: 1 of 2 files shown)


azure/functions/durable_functions.py
```diff
@@ -1,45 +1,91 @@
...
 # Durable Function Orchestration Trigger
 class OrchestrationTriggerConverter(meta.InConverter,
+                                    meta.OutConverter,
                                     binding='orchestrationTrigger',
                                     trigger=True):
     @classmethod
     def check_input_type_annotation(cls, pytype):
         return issubclass(pytype, _durable_functions.OrchestrationContext)
 
+    @classmethod
```
davidmrdavid, line 18: Why is this hardcoded and how / where is it meant to be used?

Hazhzeng: As discussed offline, I will mark this as resolved.

azure/functions/durable_functions.py
```diff
@@ -1,45 +1,91 @@
...
                data: meta.Datum, *,
-               trigger_metadata) -> Any:
-        if getattr(data, 'value', None) is not None:
-            return data.value
+               trigger_metadata) -> typing.Any:
+        data_type = data.type
+
+        # Durable functions extension always returns a string of json
+        # See durable functions library's call_activity_task docs
+        if data_type == 'string' or data_type == 'json':
```
amamounelsayed, line 63: Do we have a test to cover the decode for json type?

Hazhzeng: Yes, I add some tests for in the next iteration. Thanks for the review.

Issue #44, opened 2020-02-05
Run a type checker as part of testing


brettcannon:
To make sure the typing doesn't accidentally break.

Hazhzeng:
Hi @brettcannon,
Could you elaborate more on this one?
Do you mean running a type checker test (e.g. mypy) when we submit PR into the code base?

brettcannon:
Yep, sorry about the poorly-worded title!

PR #55, opened 2020-04-08
Fix mypy typing

Hazhzeng:
### Issue
The mypy typing check has never been properly enabled since `azure.functions` is a package instead of a module. This is now fixed in code quality test.

### Resolves
resolves https://github.com/Azure/azure-functions-python-library/issues/44

### DIFF
--- a/azure/functions/_http_wsgi.py
+++ b/azure/functions/_http_wsgi.py
@@ -10,7 +10,7 @@ from ._thirdparty.werkzeug._compat import string_types, wsgi_encoding_dance
 
 
 class WsgiRequest:
-    _environ_cache: Dict[str, Any] = None
+    _environ_cache: Optional[Dict[str, Any]] = None
 
     def __init__(self,
                  func_req: HttpRequest,
@@ -96,7 +96,7 @@ class WsgiRequest:
         if lowercased_headers.get('x-forwarded-port'):
             return int(lowercased_headers['x-forwarded-port'])
         elif getattr(parsed_url, 'port', None):
-            return parsed_url.port
+            return int(parsed_url.port)
         elif parsed_url.scheme == 'https':
             return 443
         return port
@@ -134,7 +134,7 @@ class WsgiResponse:
     # PEP 3333 start response implementation
     def _start_response(self, status: str, response_headers: List[Any]):
         self._status = status
-        self._headers = Headers(response_headers)
+        self._headers = Headers(response_headers)  # type: ignore
         self._status_code = int(self._status.split(' ')[0])  # 200 OK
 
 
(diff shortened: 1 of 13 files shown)


azure/functions/http.py
```diff
@@ -21,12 +21,14 @@ def __init__(self,
                  route_params: typing.Mapping[str, str],
                  body_type: str,
                  body: typing.Union[str, bytes]) -> None:
+
+        body_str: typing.Optional[str] = None
```
Hazhzeng, line 25: Only body_str could be None.
The request should always have body bytes. It acts as the raw body of the request itself.
The body_str should be encoded with 'utf-8' directly, rather than doing lazy encoding in get_json() method.

azure/functions/_http_wsgi.py
```diff
@@ -10,7 +10,7 @@
 
 
 class WsgiRequest:
-    _environ_cache: Dict[str, Any] = None
+    _environ_cache: Optional[Dict[str, Any]] = None
```
vrdmr, line 13: For these `Optional` - Is the goal to make them strongly typed later in the future?

azure/functions/blob.py
```diff
@@ -1,31 +1,31 @@
 import io
-import typing
+from typing import Optional, Union, Any
```
vrdmr, line 2: This is better, as we only import what's actually needed and make it explicit at the top. We should slowly be changing all the `import *` statements to `from * import **` statements.

azure/functions/http.py
```diff
@@ -46,16 +48,18 @@ def get_body(self) -> bytes:
...
-        else:
+        elif self.__body_bytes is not None:
             try:
                 return json.loads(self.__body_bytes.decode())
             except ValueError as e:
                 raise ValueError(
                     'HTTP request does not contain valid JSON data') from e
+        else:
+            raise ValueError(
+                'Request body cannot be empty in JSON deserialization') from e
```
vrdmr, line 62: Would it help if we could also return what we got in __body_bytes? Also, for `json.loads(self.__body_bytes.decode())`, do we need to make `utf-08` explicit during the decode phase?

vrdmr:
@Hazhzeng: test failure.

```
_________________________ TestCodeQuality.test_flake8 __________________________
Traceback (most recent call last):
  File "/home/vsts/work/1/s/tests/test_code_quality.py", line 51, in test_flake8
    f'flake8 validation failed:\n{output}') from None
AssertionError: flake8 validation failed:
./azure/functions/http.py:62:78: F821 undefined name 'e'
```

azure/functions/_http_wsgi.py
Hazhzeng: Nice, good catch 👍

Hazhzeng: Yes, this _environ_cache is a cache of WSGI Dict in memory. When cache miss, it will be assigned later.

Hazhzeng:
@vrdmr thanks for reviewing, I fixed the flake8 validation errors and going to merge this PR.

PR #56, opened 2020-04-22
Adding event grid's output binding and updating gitignore

vrdmr:
Adding support for event_grid's output binding support.

### DIFF
--- a/azure/functions/eventgrid.py
+++ b/azure/functions/eventgrid.py
@@ -44,14 +44,15 @@ class EventGridEventInConverter(meta.InConverter, binding='eventGridTrigger',
         )
 
 
-class EventGridOutConverter(meta.OutConverter, binding="eventGrid"):
+class EventGridEventOutConverter(meta.OutConverter, binding="eventGrid"):
     @classmethod
     def check_output_type_annotation(cls, pytype: type) -> bool:
-        valid_types = (str, bytes, azf_eventgrid.EventGridEvent,
-                       List[azf_eventgrid.EventGridEvent])
+        valid_types = (str, bytes, azf_eventgrid.EventGridOutputEvent,
+                       List[azf_eventgrid.EventGridOutputEvent])
         return (meta.is_iterable_type_annotation(pytype, str) or meta.
                 is_iterable_type_annotation(pytype,
-                azf_eventgrid.EventGridEvent) or (isinstance(pytype, type)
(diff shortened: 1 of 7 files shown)


vrdmr:
Thanks @anirudhgarg. I'll merge this once the E2E tests can be done (post the net31 issue identified in the eventgrid extension binding - https://github.com/Azure/azure-functions-eventgrid-extension/issues/82). 
@Hazhzeng Can you please take a look at the changes?

azure/functions/eventgrid.py
```diff
@@ -36,3 +43,73 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
...
+                })
+            )
+
+        elif isinstance(obj, collections.abc.Iterable):
+            msgs = []
+            for item in obj:
+                if isinstance(item, str):
+                    msgs.append(item)
+                elif isinstance(item, azf_eventgrid.EventGridEvent):
+                    msgs.append(json.dumps({'id': item.id,
```
Hazhzeng, line 91: Can we create a Python Dict here instead of using json.dumps? Since in line 105, we will dump the whole object into datum. If I'm missing something here, please let me know :)

tests/test_eventgrid.py
```diff
@@ -0,0 +1,139 @@
...
+            example_data, expected_type=type(example_data))
+        self.assertEqual(eventGridDatum.type, "string")
+
+    def test_eventgrid_encode_with_bytes_data(self):
+        example_data = self._generate_single_eventgrid_str(True)
+        eventGridDatum = azf_event_grid.EventGridOutConverter.encode(
+            example_data, expected_type=type(example_data))
+        self.assertEqual(eventGridDatum.type, "bytes")
+
+    def test_eventgrid_encode_with_EventGridData(self):
```
Hazhzeng, line 63: Is it possible to add an encode List[EventGrid] test case here? If it is not tedious, please add one.

azure/functions/eventgrid.py
vrdmr: Solved. thanks for the help.

tests/test_eventgrid.py
vrdmr: Done.

vrdmr:
Fixes https://github.com/Azure/azure-functions-python-worker/issues/638.

PR #57, opened 2020-04-22
Adds Encode and Decode logic for "arbitrary" objects for Durable Python

davidmrdavid:
This PR is a kind of "follow-up" to: https://github.com/Azure/azure-functions-python-library/pull/53

The Durable Functions Python group is looking to support custom objects as inputs and outputs for their Orchestrations and Activity Functions. To do this, we establish a kind of "contract" with the user that their custom classes will provide static `to_json` and `from_json` methods that we can use to serialize and deserialize their object instances. These functions get called as `json.dumps` and `json.loads` 'callbacks` at various points and throughout the library and the [durable-python ](https://github.com/Azure/azure-functions-durable-python) code itself.

So this PR does three things:
1) It defines the helper methods to serialize and de-serialize custom objects
2) Uses them in the ActivityTriggerConverters's `decode` and `encode` methods
3) Exposes them for further usage in Durable Python

I worry that the last point may be controversial, because we are exposing a framework-internal function and users doing `import azure.functions` would be able to see them. Now, the fact that they are prefixed by two underscores should signal that they are not for public usage. Another idea I had was to bundle the two serialization functions under a `__serialization__` module, and export that instead. Please let me know what you think, and than you 😸

### DIFF
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -19,7 +19,6 @@ from . import queue  # NoQA
 from . import servicebus  # NoQA
 from . import timer  # NoQA
 from . import durable_functions  # NoQA
-from .durable_functions import __deserialize_custom_object, __serialize_custom_object
 
 
 __all__ = (
@@ -44,11 +43,7 @@ __all__ = (
     'TimerRequest',
 
     # Middlewares
-    'WsgiMiddleware',
-
-    # serialization
-    '__serialize_custom_object'
-    '__deserialize_custom_object'
+    'WsgiMiddleware'
 )
 
 __version__ = '1.2.0'
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -9,7 +9,6 @@ from ._queue import QueueMessage  # NoQA
 from ._servicebus import ServiceBusMessage  # NoQA
 from ._durable_functions import OrchestrationContext  # NoQA
 from .meta import get_binding_registry  # NoQA
-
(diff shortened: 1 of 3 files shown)


azure/functions/durable_functions.py
```diff
@@ -80,7 +149,7 @@ def decode(cls,
     def encode(cls, obj: typing.Any, *,
                expected_type: typing.Optional[type]) -> meta.Datum:
         try:
-            result = json.dumps(obj)
+        result = json.dumps(obj, default=__serialize_custom_object)
```
davidmrdavid, line 152: whoops, the indentation here is wrong. I'll fix it shortly

azure/functions/__init__.py
```diff
@@ -45,6 +45,10 @@
 
     # Middlewares
     'WsgiMiddleware',
+
+    # serialization
+    '__serialize_custom_object'
+    '__deserialize_custom_object'
```
Hazhzeng, line 51: This is the export statement of `azure.functions`. I don't think we actually need these two lines here since the user of our library would not use `func.__serialize_custom_object` or `func.__deserialize_custom_object`.

Hazhzeng:
Since these custom_objects are very durable function specific, I would suggest just let our user to do `import azure.functions.durable_functions.__serialize_custom_object` instead

azure/functions/durable_functions.py
davidmrdavid: Fxied

PR #58, opened 2020-04-30
Adding new task in CI to report code-coverage in branches and PRs.

vrdmr:


.coveragerc
```diff
@@ -0,0 +1,41 @@
...
+    if self\.debug
+
+    # Don't complain if tests don't hit defensive assertion code:
+    raise AssertionError
+    raise NotImplementedError
+
+    # Don't complain if non-runnable code isn't run:
+    if 0:
+    if False:
+    if __name__ == .__main__.:
```
Hazhzeng, line 18: Just curious, what's `.__main__.` stands for? It this dot `.` here a wildcard?

vrdmr: I picked up the config from the [configuration documentation](https://coverage.readthedocs.io/en/latest/config.html#syntax), and it is a regex wildcard.

Some more info about `exclude_lines` section.
> exclude_lines (multi-string): a list of regular expressions. Any line of your source code that matches one of these regexes is excluded from being reported as missing. More details are in [Excluding code from coverage.py](https://coverage.readthedocs.io/en/latest/excluding.html#excluding). If you use this option, you are replacing all the exclude regexes, so you’ll need to also supply the “pragma: no cover” regex if you still want to use it.

PR #65, opened 2020-07-10
Fix EnqueueTimeUtc in EventHub and expose ServiceBus new properties 

Hazhzeng:
### Descriptions
1. The single EventHub enqueue_time field now extract data from `trigger_metadata['EnqueuedTimeUtc']` instead of `trigger_metadata['EnqueuedTime']`.
2. The newly introduced ServiceBus properties **expires_at_utc** and **enqueued_time_utc** need to be exposed at _servicebus.py public interface.
3. Add more robust tests in EventHub event properties and ServiceBus message properties.

### DIFF
--- a/azure/functions/eventhub.py
+++ b/azure/functions/eventhub.py
@@ -125,7 +125,7 @@ class EventHubTriggerConverter(EventHubConverter,
             body=body,
             trigger_metadata=trigger_metadata,
             enqueued_time=cls._parse_datetime_metadata(
-                trigger_metadata, 'EnqueuedTime'),
+                trigger_metadata, 'EnqueuedTimeUtc'),
             partition_key=cls._decode_trigger_metadata_field(
                 trigger_metadata, 'PartitionKey', python_type=str),
             sequence_number=cls._decode_trigger_metadata_field(
--- a/tests/test_servicebus.py
+++ b/tests/test_servicebus.py
@@ -91,23 +91,21 @@ class TestServiceBus(unittest.TestCase):
 
         # Datetime should be in iso8601 string instead of datetime object
         metadata_dict = servicebus_msg.metadata
-        self.assertEqual(metadata_dict['DeliveryCount'], 1)
-        self.assertEqual(metadata_dict['LockToken'],
-                         '87931fd2-39f4-415a-9fdc-adfdcbed3148')
-        self.assertEqual(metadata_dict['ExpiresAtUtc'],
-                         '2020-07-02T05:39:12.17Z')
-        self.assertEqual(metadata_dict['EnqueuedTimeUtc'],
-                         self.MOCKED_ENQUEUE_TIME.isoformat())
-        self.assertEqual(metadata_dict['MessageId'],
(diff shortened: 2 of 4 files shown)


azure/functions/eventhub.py
```diff
@@ -125,7 +125,7 @@ def decode_single_event(
             body=body,
             trigger_metadata=trigger_metadata,
             enqueued_time=cls._parse_datetime_metadata(
-                trigger_metadata, 'EnqueuedTime'),
```
vrdmr, line 128: So are we making a change from `EnqueuedTime` to `EnqueuedTimeUTC`? Could that break existing users of the library which depend on that value to be in local time?

vrdmr: Or was it never available in local time?

tests/test_servicebus.py
```diff
@@ -135,14 +147,53 @@ def _generate_servicebus_metadata(self):
...
+        mocked_metadata['PartitionKey'] = meta.Datum('sample_part', 'string')
         mocked_metadata['Label'] = meta.Datum(
             'Microsoft.Azure.ServiceBus', 'string'
         )
-        mocked_metadata['sys'] = meta.Datum(type='json', value='''
-            {
-                "MethodName": "ServiceBusSMany",
-                "UtcNow": "2020-06-18T05:39:12.2860411Z",
-                "RandGuid": "bb38deae-cc75-49f2-89f5-96ec6eb857db"
+        mocked_metadata['MessageReceiver'] = meta.Datum(type='json', value='''
```
vrdmr, line 154: This is the new section of metadata received?

tests/test_servicebus.py
```diff
@@ -79,7 +91,7 @@ def test_servicebus_metadata(self):
 
         # Datetime should be in iso8601 string instead of datetime object
         metadata_dict = servicebus_msg.metadata
-        self.assertDictEqual(metadata_dict, {
+        self.assertGreaterEqual(metadata_dict.items(), {
```
vrdmr, line 94: Does this compare all the metadata items? If yes, then why the "Greater"?

azure/functions/eventhub.py
Hazhzeng: > Or was it never available in local time?

It never works in local time. The current implementation will always return None to our customer.

tests/test_servicebus.py
Hazhzeng: Yes, this is the new section of metadata provided by the extension. We still support the old **mocked_metadata['sys']** in line 192.

tests/test_servicebus.py
Hazhzeng: This test is amended to ensure the metadata_dict has the following expected fields (**DeliveryCount**, **LockToken**, **ExpiresAtUtc**...).

Using **self.assertDictEqual** is too strict for this test scenario. We just want to ensure the expected fields are a subset of the metadata. I tried using **self.assertDictContainsSubset** but it has been deprecated in Python 3.6. So I first convert the dictionary into an ItemView using **.items()**, and use **self.assertGreaterEqual** for subset check.

PR #68, opened 2020-08-25
Added support for Entity Trigger

davidmrdavid:
The Durable Entities construct of Durable Functions uses a trigger binding type named `entityTrigger`. This PR makes a minimal change that enables `entityTrigger` to enable Durable Functions for Python to define Entities. This is will enable the following PR for Durable Functions in Python: https://github.com/Azure/azure-functions-durable-python/pull/184

### DIFF
--- a/azure/functions/durable_functions.py
+++ b/azure/functions/durable_functions.py
@@ -38,6 +38,7 @@ class OrchestrationTriggerConverter(meta.InConverter,
     def has_implicit_output(cls) -> bool:
         return True
 
+
 class EnitityTriggerConverter(meta.InConverter,
                               meta.OutConverter,
                               binding='entityTrigger',
--- a/azure/functions/durable_functions.py
+++ b/azure/functions/durable_functions.py
@@ -38,6 +38,34 @@ class OrchestrationTriggerConverter(meta.InConverter,
     def has_implicit_output(cls) -> bool:
         return True
 
+class EnitityTriggerConverter(meta.InConverter,
+                                    meta.OutConverter,
+                                    binding='entityTrigger',
+                                    trigger=True):
+    @classmethod
+    def check_input_type_annotation(cls, pytype):
+        return issubclass(pytype, _durable_functions.EntityContext)
+
+    @classmethod
+    def check_output_type_annotation(cls, pytype):
+        # Implicit output should accept any return type
+        return True
+
+    @classmethod
+    def decode(cls,
+               data: meta.Datum, *,
+               trigger_metadata) -> _durable_functions.EntityContext:
+        return _durable_functions.EntityContext(data.value)
+
+    @classmethod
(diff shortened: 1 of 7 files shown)


azure/functions/durable_functions.py
```diff
@@ -39,6 +39,36 @@ def has_implicit_output(cls) -> bool:
...
+    @classmethod
+    def check_output_type_annotation(cls, pytype):
+        # Implicit output should accept any return type
+        return True
+
+    @classmethod
+    def decode(cls,
+               data: meta.Datum, *,
+               trigger_metadata) -> _durable_functions.EntityContext:
+        return _durable_functions.EntityContext(data.value)
```
Hazhzeng, line 59: Could it be possible that the **data** from the EntityContext trigger is set to None? We skip the None check in OrchestrationContext since we know it never returns None. Just want you to double check 👍 Also, I would like to know more on the EntityContext use case.

Hazhzeng:
Please also importing this EntityContext in `azure/functions/__init__.py`

vrdmr:
@davidmrdavid - Please add tests for the new type in here - [test_durable_functions.py](https://github.com/Azure/azure-functions-python-library/blob/dev/tests/test_durable_functions.py). Also, when this gets done, also add to the E2E tests in worker - the second piece will come in later but just FYI.

vrdmr:
@davidmrdavid Any updates on this?

davidmrdavid:
@vrdmr , I haven't gotten around this yet because I was OOF last week and got a bit sidetracked with some other tasks today. I want to make this a priority for tomorrow though. I will be pinging you for a few clarifications on this in the morning, sorry for the delay! Just for my information, is this blocking a release?

azure/functions/durable_functions.py
davidmrdavid: As per our discussion, I'll keep this as-is. I don't believe `data` can be set to None, as it should be very much like the OrchestrationContext

davidmrdavid:
@Hazhzeng @vrdmr ,
Made the requested changes (exporting the EntityContext) and **generalized** the existing unit tests to account for the EntityContext as well.

@Hazhzeng , I also left a TODO on the existing unit tests as I realized that [these two tests](https://github.com/Azure/azure-functions-python-library/blob/18368cb85174a0372db47c349941f8d3a784497a/tests/test_durable_functions.py#L32-L46) are identical. Please let me know if I'm missing something there.

azure/functions/durable_functions.py
vrdmr: For this, let's have an offline discussion.

> Also, I would like to know more on the EntityContext use case.

vrdmr (approved):
Thanks for the changes - Let's have a quick session on how the EntityTrigger is used.

Hazhzeng (approved):

PR #77, opened 2020-12-08
Exposing properties and metadata fields for blobs

vrdmr:
This PR exposes the two remaining fields - Blob Properties and Metadata to be accessed.


Hazhzeng (approved):
Thanks for cleaning up the readme.md. This PR looks good to me!

Issue #71, opened 2020-09-24
annotations for WsgiMiddleware cause app to not be startable


asottile:
```console
$ ./node_modules/.bin/func host start --python --functions MyFunction
Found Python version 3.8.2 (python3).
Azure Functions Core Tools (3.0.2912 Commit hash: bfcbbe48ed6fdacdf9b309261ecc8093df3b83f2)
Function Runtime Version: 3.0.14287.0
Hosting environment: Production
Content root path: /home/asottile/workspace/pre-commit-ci/azure-sample-application
Now listening on: http://0.0.0.0:7071
Application started. Press Ctrl+C to shut down.

Functions:

	MyFunction: [GET,POST] http://localhost:7071/api/MyFunction

For detailed output, run func with --verbose flag.
Worker process started and initialized.
Worker failed to function id 6333f1be-d5f7-4f0d-8cb8-e8ed6f42210a.
Result: Failure
Exception: FunctionLoadError: cannot load the MyFunction function: the "context" parameter is expected to be of type azure.functions.Context, got typing.Union[azure.functions._abc.Context, NoneType]
Stack:   File "/home/asottile/workspace/pre-commit-ci/azure-sample-application/node_modules/azure-functions-core-tools/bin/workers/python/3.8/LINUX/X64/azure_functions_worker/dispatcher.py", line 268, in _handle__function_load_request
    self._functions.add_function(
  File "/home/asottile/workspace/pre-commit-ci/azure-sample-application/node_modules/azure-functions-core-tools/bin/workers/python/3.8/LINUX/X64/azure_functions_worker/functions.py", line 105, in add_function
    raise FunctionLoadError(
```

```python
import azure.functions
import flask

app = flask.Flask(__name__)


@app.route('/', methods=['GET'])
def index() -> str:
    return 'hello hello world'


main = azure.functions.WsgiMiddleware(app).main
```

it looks like it is being a little too strict on the typing here

the `@property` is annotated correctly [here](https://github.com/Azure/azure-functions-python-library/blob/e8f74bc8bc447e15e7268a508ffcdd731a357075/azure/functions/_http_wsgi.py#L152)

but the function it returns, uses `Optional[Context]` instead [here](https://github.com/Azure/azure-functions-python-library/blob/e8f74bc8bc447e15e7268a508ffcdd731a357075/azure/functions/_http_wsgi.py#L159)

[...]

PR #88, opened 2021-05-03
Add message when ASGI and WSGI is enabled

Hazhzeng:
### Related Issues
Fix: https://github.com/Azure/azure-functions-python-library/issues/71
Fix: https://github.com/Azure/azure-functions-python-worker/issues/839

### Description
1. Now when the WSGI and ASGI middleware start, we will emit `Instantiating Azure Functions ASGI middleware.` or `Instantiating Azure Functions WSGI middleware.` log message for the very first time when the middleware instantiated either in `main = WsgiMiddleware(app)` syntax or `def func(req, context): return WsgiMiddleware(app).handle(req, context)` syntax.
2. Fix an issue where the middlewares fail to accept the context parameter.

### Todo
1. In the worker, we need to direct `azure.functions` log into system log category.

cc: @tonybaloney @stefanushinardi @vrdmr

### DIFF
--- a/tests/test_http_asgi.py
+++ b/tests/test_http_asgi.py
@@ -153,3 +153,18 @@ class TestHttpAsgiMiddleware(unittest.TestCase):
         # Verify asserted
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.get_body(), test_body)
+
+    def test_middleware_wrapper(self):
+        app = MockAsgiApplication()
+        test_body = b'Hello world!'
+        app.response_body = test_body
+        app.response_code = 200
+        req = self._generate_func_request()
+        ctx = self._generate_func_context()
+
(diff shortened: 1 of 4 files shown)


tests/test_http_asgi.py
```diff
@@ -153,3 +153,18 @@ def test_middleware_calls_app_with_context(self):
         # Verify asserted
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.get_body(), test_body)
+
+    def test_middleware_wrapper(self):
```
vrdmr, line 157: PyDoc again - what this is testing. And for below as well.

vrdmr (approved):
LGTM 🚢 

nit - comments wherever needed.

stefanushinardi (approved):

PR #95, opened 2021-10-11
Add missing date/time formats for Servicebus

davidmrdavid:
This is a working patch for: https://github.com/Azure/azure-functions-python-worker/issues/897

In short, it appears that the Python worker does not support some date and timedelta string formats that ServiceBus supports. This PR is the result of looking at the runtime data sent by the reproducer for the IcM referenced in the link above and manually adding in the missing date/timedelta string representations. It would be good to get the opinion of our NodeJS worker maintainers to understand the following:

(1) Are we still missing any other date time / timedelta formats compared to the node worker?
(2) **Where** does the node worker handle the parsing of these dates? I couldn't find that information myself.

After getting the answers to the questions above, we can look into adding a few test cases to this PR. Thanks!

### DIFF
--- a/azure/functions/meta.py
+++ b/azure/functions/meta.py
@@ -360,7 +360,6 @@ class _BaseConverter(metaclass=_ConverterMeta, binding=None):
         """
 
         timedelta_formats = [
-            '%d.%H:%M:%S',
             '%H:%M:%S',
             '%M:%S',
             '%S'


vrdmr:
> (1) Are we still missing any other date time / timedelta formats compared to the node worker?

This was the only parsing error that showed up in our telemetry. I have added some more formats to be future-safe.

> (2) Where does the node worker handle the parsing of these dates? I couldn't find that information myself.

Node doesn't do a parsing explicitly - it leverages the JSON parsing to parse the incoming gRPC object and doesn't do any explicit parsing as such.

PR #97, opened 2021-11-18
Exclude azure from package build

gavin-aguiar:
### Description:
Excluded azure from the packages. This fixes an issue where the azure functions version does not get updated when a version is specified in the functions requirements.txt file.
Reason for this is since azure directory has an __init.py__ file, including azure in the package build overrides the users dependencies.


vrdmr (approved):
Thanks for the changes. LGTM

PR #100, opened 2022-02-08
Adding Kafka headers to binding

shrohilla:
Adding support for kafka headers :-

- Headers Kafka event should be able to read headers from metadata


vrdmr (approved):
LGTM.

PR #105, opened 2022-03-08
Added decorator for storage blobs

gavin-aguiar:
Added new storage blob decorator for the new programming model


azure/functions/decorators/blob.py
```diff
@@ -0,0 +1,50 @@
+from typing import Optional
```
vrdmr, line 1: Copyright Comment

vrdmr, line 1: Change the module name to _blob - potential conflict possibility.

azure/functions/decorators/function_app.py
```diff
@@ -1071,3 +1073,134 @@ def decorator():
             return decorator()
 
         return wrap
+
+    def on_blob_change(self,
+                       arg_name: str,
```
vrdmr, line 1078: Is there a plan for adding a default value for arg_name? @YunchuWang - did we add default arg name values for other triggers/bindings?

vrdmr (approved):
only small comments. Otherwise LGTM.

tests/decorators/test_blob.py
```diff
@@ -0,0 +1,57 @@
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License.
+import unittest
+
+from azure.functions import DataType
```
vrdmr, line 5: nit - directly import from the defining module and not from the top-level module. We the path changes in the future, this should be caught.

azure/functions/decorators/function_app.py
gavin-aguiar: Arg name does not have a default value.. We can definitely add a default value later one we have our templates ready.

azure/functions/decorators/blob.py
gavin-aguiar: Fixed

gavin-aguiar: As discussed, keeping the module name as blob since it is not exposed and wont cause a conflict.

tests/decorators/test_blob.py
```diff
@@ -0,0 +1,56 @@
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License.
+import unittest
+
+from azure.functions.decorators.blob import BlobTrigger, BlobOutput, BlobInput
+from azure.functions.decorators.core import BindingDirection, DataType
+
+
+class TestBlob(unittest.TestCase):
+    def test_blob_trigger_valid_creation(self):
```
vrdmr, line 10: Can you also add a small test where the decorator is actually getting assigned?

vrdmr: Essentially - a new blob decorator test in `test_decorators.py`.

vrdmr: This would address CodeCov check as well - https://github.com/Azure/azure-functions-python-library/pull/105/checks?check_run_id=5488935060

YunchuWang: ditto, everything else looks good

YunchuWang (approved):

vrdmr (approved):
:shipit:

PR #110, opened 2022-03-30
Add missing fields in timer object.

pdthummar:
Added missing fields in timer object. 
Fixes: https://github.com/Azure/azure-functions-python-worker/issues/681
Fixes: https://github.com/Azure/azure-functions-python-worker/issues/858

Missing fields

### DIFF
--- a/azure/functions/timer.py
+++ b/azure/functions/timer.py
@@ -1,6 +1,3 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License.
-
 import json
 import typing
 
@@ -11,13 +8,22 @@ from . import meta
 
 class TimerRequest(azf_abc.TimerRequest):
 
-    def __init__(self, *, past_due: bool) -> None:
+    def __init__(self, *, past_due: bool, schedulestatus:dict , schedule:dict) -> None:
         self.__past_due = past_due
+        self.__schedulestatus = schedulestatus
+        self.__schedule = schedule
 
     @property
     def past_due(self):
         return self.__past_due
 
+    @property
+    def schedulestatus(self):
+        return self.__schedulestatus
+
+    @property
+    def schedule(self):
+        return self.__schedule
 
 class TimerRequestConverter(meta.InConverter,
                             binding='timerTrigger', trigger=True):
@@ -32,5 +38,8 @@ class TimerRequestConverter(meta.InConverter,
             raise NotImplementedError
 
         info = json.loads(data.value)
+
         return TimerRequest(
-            past_due=info.get('IsPastDue', False))
+            past_due=info.get('IsPastDue', False),
+            schedulestatus=info["ScheduleStatus"],
+            schedule=info["Schedule"])
--- a/azure/functions/timer.py
+++ b/azure/functions/timer.py
@@ -8,7 +8,7 @@ from . import meta
 
 class TimerRequest(azf_abc.TimerRequest):
 
-    def __init__(self, *, past_due: bool, schedulestatus:dict , schedule:dict) -> None:
+    def __init__(self, *, past_due: bool, schedulestatus: dict, schedule: dict) -> None:
         self.__past_due = past_due
         self.__schedulestatus = schedulestatus
         self.__schedule = schedule
@@ -24,6 +24,7 @@ class TimerRequest(azf_abc.TimerRequest):
     @property
     def schedule(self):
         return self.__schedule
+
 
 class TimerRequestConverter(meta.InConverter,
                             binding='timerTrigger', trigger=True):
--- a/azure/functions/timer.py
+++ b/azure/functions/timer.py
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
 import json
 import typing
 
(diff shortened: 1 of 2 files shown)


azure/functions/timer.py
```diff
@@ -1,23 +1,30 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License.
```
gavin-aguiar, line 2: Why is this removed?

azure/functions/timer.py
```diff
@@ -32,5 +39,8 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
             raise NotImplementedError
 
         info = json.loads(data.value)
+
         return TimerRequest(
-            past_due=info.get('IsPastDue', False))
+            past_due=info.get('IsPastDue', False),
+            schedulestatus=info["ScheduleStatus"],
```
gavin-aguiar, line 45: We should probably keep this the same as how we retrieve pastdue with a default value.  e.g info.get("ScheduleStatus" , {})

azure/functions/timer.py
```diff
@@ -1,23 +1,30 @@
...
         self.__past_due = past_due
+        self.__schedulestatus = schedulestatus
+        self.__schedule = schedule
 
     @property
     def past_due(self):
         return self.__past_due
 
+    @property
+    def schedulestatus(self):
```
gavin-aguiar, line 21: snake case this, schedule_status

tests/test_timer.py
```diff
@@ -0,0 +1,31 @@
...
+
+import json
+import unittest
+
+import azure.functions.timer as timer
+from azure.functions.meta import Datum
+
+
+class TestTimer(unittest.TestCase):
+    def test_TimerRequestConverter_decode(self):
```
gavin-aguiar, line 12: We might need to add some more tests. Check the test cases in other triggers. 
e.g. test_timer_input_type
test_timer_output_type
test_timer_decode
test_timer_initialize_without_args
test_timer_initialize_all_args

gavin-aguiar (changes requested):

azure/functions/timer.py
```diff
@@ -1,23 +1,30 @@
...
+        return self.__schedulestatus
+
+    @property
+    def schedule(self):
```
vrdmr, line 25: Can we also write the expected return types for all the properties?

azure/functions/timer.py
vrdmr: Yes - what if info doesn't have ScheduleStatus - it would throw key error. - we should take care of this.

vrdmr (changes requested):

azure/functions/timer.py
pdthummar: It was mistakenly removed. I have added in new commit.

azure/functions/timer.py
pdthummar: updated this as per recommendation.

azure/functions/timer.py
pdthummar: I have updated this change as per recommendation.

azure/functions/timer.py
pdthummar: I have included expected return types in updated file. Thanks.

tests/test_timer.py
pdthummar: I have added few more tests. thanks

azure/functions/timer.py
```diff
@@ -5,19 +5,29 @@
...
-    def __init__(self, *, past_due: bool) -> None:
+    def __init__(self, *, past_due: bool = False, schedule_status: dict = {},
+                 schedule: dict = {}) -> None:
         self.__past_due = past_due
+        self.__schedule_status = schedule_status
+        self.__schedule = schedule
 
     @property
-    def past_due(self):
+    def past_due(self) -> bool:
```
vrdmr, line 20: 😍  Thanks a lot for this.

azure/functions/timer.py
```diff
@@ -5,19 +5,29 @@
 import typing
 
 from azure.functions import _abc as azf_abc
-
 from . import meta
 
 
 class TimerRequest(azf_abc.TimerRequest):
```
vrdmr, line 11: @YunchuWang - Getting your input here - If we plan to add PyDocs here - what could we add here?

@pdthummar No need to worry in this PR (mostly). But we should explore adding more PyDoc comments as they eventually show up in [azure.functions.timerrequest](https://docs.microsoft.com/en-us/python/api/azure-functions/azure.functions.timerrequest?view=azure-python)

tests/test_timer.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
```
vrdmr, line 2: Curious - weren't there any Timer tests before this? 😱

tests/test_timer.py
```diff
@@ -0,0 +1,66 @@
...
+import json
+import unittest
+
+import azure.functions.timer as timer
+from azure.functions.meta import Datum
+
+
+class TestTimer(unittest.TestCase):
+    def test_timer_decode(self):
+        # given
```
vrdmr, line 13: I liked the given-when-then structure you used. 😃  👍🏼

vrdmr (approved):
🚢  
LGTM

tests/test_timer.py
pdthummar: there were not any timer tests before.

gavin-aguiar (approved):
LGTM

azure/functions/timer.py
YunchuWang: @vrdmr  would suggest giving a overview what this class does, adding definition of each field in init, and link to https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer?tabs=in-process&pivots=programming-language-java#attributes

pdthummar:
@lak-amith531 If you are using old Coretools version, Can you please upgrade it to latest version and try it? It should work with latest version. 

If you are still facing issue after upgrading version, please create an issue with including replication steps. Thanks

tim-sparq:
this doesn't work when schedule is more frequent than 1x p/min, even when useMonitor is set to true in function.json

Issue #107, opened 2022-01-16
Python - azure.functions.WsgiMiddleware does not support multiple Set-Cookie response headers


bittebak:
# Bug
When a HttpResponse has multiple headers with the same name, only one header is returned.
Django uses multiple response cookies  (e.g. 'csrf' and 'messages').
Multiple Set-Cookie headers can be returned. Set-Cookie is the name (the key) of the Http Response Header.

# Context
A Django app running in an Azure Function

## package
```
Name: azure-functions
Version: 1.8.0
Summary: Azure Functions for Python
Home-page: UNKNOWN
Author: Microsoft Corporation
Author-email: azpysdkhelp@microsoft.com
```

## Version
Azure function runtime version = ~4

## Reproduction
1. Create a http response with two Set-Cookie headers
This header has been captured from the app running without the Azure wsgi middleware.

```
HTTP/1.1 200 OK
Date: Sun, 16 Jan 2022 09:00:59 GMT
Server: WSGIServer/0.2 CPython/3.9.9
Content-Type: text/html; charset=utf-8
X-Frame-Options: DENY
Vary: Cookie
Content-Length: 5206
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Set-Cookie:  messages=""; expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; Max-Age=0; Path=/; SameSite=Lax
Set-Cookie:  csrftoken=L6coOZOclDhAg2X7geBKbhvJSHPxz9AbgOkoVqNAIsgVe8oxTEPTSPbc2VPSyjwq; expires=Sun, 15 Jan 2023 09:00:59 GMT; Max-Age=31449600; Path=/; SameSite=Lax
```

2. Run without azure.functions.WsgiMiddleware
Two  Set-Cookies headers are returned (as depicted above)

3. Run with azure.functions.WsgiMiddleware.handle
```
import azure.functions as func
def main(req: func.HttpRequest, context: func.Context) -> func.HttpResponse:
    http_response = func.WsgiMiddleware(application).handle(req, context)
    return http_response
```

Only one header with the name "Set-Cookie" is returned.

### Local and remote
This behaviour has been reproduced both locally (while debugging) as on an Azure Function container (icarus-int)

# Root cause
Python package 'azure-functions'
azure\functions\_http.py

```
class HttpResponseHeaders(BaseHeaders, collections.abc.MutableMapping):

    def __setitem__(self, key: str, value: str):
[...]

PR #115, opened 2022-04-07
Fix for GH#107: Enable support for multiple Set-Cookie response headers (native & wsgi/asgi)

YunchuWang:
Part of change to address https://github.com/Azure/azure-functions-python-library/issues/107
Related change in worker: https://github.com/Azure/azure-functions-python-worker/pull/1004

### Changes
1. Make HttpResponseHeader inherit from werkzeug header class which is also used in Flask, the header class has dict-like interface but is ordered and can store the same keys multiple times, all operations are case insensitive on header key.
2. Cookie headers are extracted from httpresponseheaders and sent to worker as cookie field in rpchttp message.

### DIFF
--- a/azure/functions/_abc.py
+++ b/azure/functions/_abc.py
@@ -193,7 +193,7 @@ class HttpResponse(abc.ABC):
 
     @property
     @abc.abstractmethod
-    def headers(self) -> typing.MutableMapping[str, str]:
+    def headers(self):
         pass
 
     @abc.abstractmethod
--- a/azure/functions/_abc.py
+++ b/azure/functions/_abc.py
@@ -6,6 +6,7 @@ import datetime
 import io
 import typing
 
+from azure.functions._thirdparty.werkzeug.datastructures import Headers
 
 T = typing.TypeVar('T')
 
@@ -193,7 +194,7 @@ class HttpResponse(abc.ABC):
 
     @property
     @abc.abstractmethod
-    def headers(self):
+    def headers(self) -> Headers:
         pass
 
     @abc.abstractmethod
--- a/azure/functions/_http.py
+++ b/azure/functions/_http.py
@@ -4,14 +4,14 @@
 import collections.abc
 import io
 import json
-import typing
 import types
+import typing
 
 from . import _abc
-
 from ._thirdparty.werkzeug import datastructures as _wk_datastructures
 from ._thirdparty.werkzeug import formparser as _wk_parser
 from ._thirdparty.werkzeug import http as _wk_http
+from ._thirdparty.werkzeug.datastructures import Headers
 
 
 class BaseHeaders(collections.abc.Mapping):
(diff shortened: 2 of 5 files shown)


azure/functions/http.py
```diff
@@ -6,9 +6,13 @@
 
 from azure.functions import _abc as azf_abc
 from azure.functions import _http as azf_http
-
 from . import meta
 
+try:
+    from http.cookies import SimpleCookie
+except ImportError:
+    from Cookie import SimpleCookie  # type: ignore
```
vrdmr, line 14: Which `http` module is this? I couldn't find he definition for SimpleCookie. Also, for Cookie as well.

azure/functions/http.py
```diff
@@ -79,7 +83,8 @@ def encode(cls, obj: typing.Any, *,
 
         if isinstance(obj, azf_abc.HttpResponse):
             status = obj.status_code
-            headers = dict(obj.headers)
+            headers = obj.headers
```
vrdmr, line 86: This needs to be type-ified to help read the code later.

vrdmr (approved):
LGTM.

Left some comments and we can discuss offline to go over the details.

azure/functions/_abc.py
```diff
@@ -193,7 +193,7 @@ def charset(self):
 
     @property
     @abc.abstractmethod
-    def headers(self) -> typing.MutableMapping[str, str]:
+    def headers(self):
```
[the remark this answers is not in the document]
YunchuWang: will update

azure/functions/_http.py
```diff
@@ -40,13 +40,8 @@ class HttpRequestHeaders(BaseHeaders):
...
 
 
-class HttpResponseHeaders(BaseHeaders, collections.abc.MutableMapping):
-
-    def __setitem__(self, key: str, value: str):
-        self.__http_headers__[key.lower()] = value
-
-    def __delitem__(self, key: str):
-        del self.__http_headers__[key.lower()]
+class HttpResponseHeaders(Headers):
```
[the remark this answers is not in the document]
YunchuWang: yes, its only for request headers. for response headers only, we need to be able to set multi cookie headers which needs it to be extending werkzeug headers class.

azure/functions/http.py
YunchuWang: its the standard http python module (https://docs.python.org/3.8/library/http.cookies.html). For Cookie module it is a old module in python 2.7 https://docs.python.org/2/library/cookie.html. 

image

Good point, python 2 is not supported, i can remove the except block.

azure/functions/http.py
YunchuWang: will update

azure/functions/http.py
```diff
@@ -93,6 +98,12 @@ def encode(cls, obj: typing.Any, *,
             else:
                 datum_body = meta.Datum(type='bytes', value=b'')
 
+            cookies = None
+            if "Set-Cookie" in headers:
```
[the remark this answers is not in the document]
YunchuWang: no, we need to convert them to RPChttpCookie at https://github.com/Azure/azure-functions-python-worker/blob/f4694d001dc762b9ca8fb06bfcfae098b2fef235/azure_functions_worker/bindings/datumdef.py#L188-L197

PR #116, opened 2022-04-11
Blueprint interfaces implementation

YunchuWang:
### Introducing blueprint concepts to new programming model (referenced to flask blueprint)

**The benefits of blueprint is:**
- separating functions to different python files
- providing extensible, module python classes to enable external parties to implement their own function app types.

**New public classes blueprint exposes:**
- DecoratorAPI: An abstract class which contains essential decorator function building blocks to extend for creating new function app or blueprint classes.
- HttpFunctionsAuthLevelMixin: Python class which manages(getting/setting) http authorization level property to the extended classes.
- TriggerApi: Interface to extend for using existing trigger decorator functions.
- BindingApi: Interface to extend for using existing binding decorator functions.
- FunctionRegister: Interface for declaring top level function app class which will be directly indexed by Python Function runtime. There can only be one FunctionRegister instance in a function app.
- FunctionApp: FunctionApp instance containing all natively supported decorator functions to capture user defined function and corresponding metadata information.
- Blueprint: Similar to FunctionApp with only difference is BluePrint does not extends FunctionRegister class.
- ExternalHttpFunctionApp: Interface to extend for building third party http function apps.
- AsgiFunctionApp: Top level function app class for constructing asgi http functions.
- WsgiFunctionApp: Top level function app class for constructing wsgi http functions.

image


azure/functions/decorators/function_app.py
```diff
@@ -178,50 +179,29 @@ def _validate_function(self) -> None:
                 f" in bindings {bindings}")
 
         # Set route to function name if unspecified in the http trigger
```
gavin-aguiar, line 181: Can you update this comment to mention about setting the auth_level?

azure/functions/decorators/function_app.py
```diff
@@ -302,22 +264,28 @@ def decorator():
...
 
-        :param http_middleware: :class:`AsgiMiddleware` or
-        :class:`WsgiMiddleware` instance.
+class HttpFunctionsAuthLevelMixin(ABC):
+    """Interface to extend for enabling function app level http
+    authorization level setting"""
 
-        :return: None
+    def __init__(self, auth_level: Union[AuthLevel, str], *args, **kwargs):
+        self._auth_level = AuthLevel[auth_level] \
```
gavin-aguiar, line 273: What if auth_level is None?

azure/functions/decorators/function_app.py
```diff
@@ -302,22 +264,28 @@ def decorator():
...
         return wrap
 
-    def _add_http_app(self,
-                      http_middleware: Union[
-                          AsgiMiddleware, WsgiMiddleware]) -> None:
-        """Add a Wsgi or Asgi app integrated http function.
 
-        :param http_middleware: :class:`AsgiMiddleware` or
-        :class:`WsgiMiddleware` instance.
+class HttpFunctionsAuthLevelMixin(ABC):
```
gavin-aguiar, line 268: What is the Mixin at the end of the classname? Can we rename to IHttpFunctionsAuthLevel . Similar comment to the other interfaces. We can add an 'I' before the class name to differentiate interfaces from other classes

azure/functions/decorators/function_app.py
```diff
@@ -302,22 +264,28 @@ def decorator():
...
 
-        @self.route(methods=(method for method in HttpMethod),
-                    auth_level=self.auth_level,
-                    route="/{*route}")
-        def http_app_func(req: HttpRequest, context: Context):
-            return http_middleware.handle(req, context)
+        return self._auth_level
+
+
+class TriggerApi(DecoratorApi, ABC):
```
gavin-aguiar, line 287: Similar comment here. We can change this to ITriggerApi

azure/functions/decorators/function_app.py
```diff
@@ -302,22 +264,28 @@ def decorator():
...
+    """Interface to extend for enabling function app level http
+    authorization level setting"""
 
-        :return: None
+    def __init__(self, auth_level: Union[AuthLevel, str], *args, **kwargs):
+        self._auth_level = AuthLevel[auth_level] \
+            if isinstance(auth_level, str) else auth_level
+
+    @property
+    def auth_level(self):
```
gavin-aguiar, line 277: Return typing

azure/functions/decorators/function_app.py
```diff
@@ -1324,52 +1359,104 @@ def decorator():
...
+                in self._function_builders]
 
-        :return: Decorator function.
+    def register_functions(self, functions: DecoratorApi) -> None:
+        """Register a list of functions in the function app.
+
+        :param functions: Instance extending :class:`DecoratorApi` which
+        contains a list of functions.
         """
+        if isinstance(functions, FunctionRegister):
```
gavin-aguiar, line 1390: functions is a list right? How are we checking isinstance of a list to a FunctionResgister object?

YunchuWang: need better naming, its not a list but a functions container instance

azure/functions/decorators/function_app.py
YunchuWang: then authlevel will not be included in binding.json and host will set default auth level to Function

azure/functions/decorators/function_app.py
YunchuWang: Mixin is implying this is not is-a relationship, https://www.pythontutorial.net/python-oop/python-mixin/#:~:text=What%20is%20a%20mixin%20in%20Python%20A%20mixin,mixin%20bundles%20a%20set%20of%20methods%20for%20reuse.

For I prefix, I do not see any other abstract classes in library use it. Lets keep it consistent. Also in python standard libraries do not see a pattern of using I prefix
image

azure/functions/decorators/function_app.py
```diff
@@ -1326,140 +1463,104 @@ def decorator():
...
+                in self._function_builders]
 
-        :return: Decorator function.
+    def register_functions(self, function_container: DecoratorApi) -> None:
+        """Register a list of functions in the function app.
+
+        :param function_container: Instance extending :class:`DecoratorApi`
+        which contains a list of functions.
         """
+        if isinstance(function_container, FunctionRegister):
```
vrdmr, line 1494: Why is this explicit check being added? In what case are you seeing this?

azure/functions/decorators/function_app.py
```diff
@@ -1326,140 +1463,104 @@ def decorator():
...
-         parameter value.
-        :param kwargs: Keyword arguments for specifying additional binding
-        fields to include in the binding json.
+        :return: List of functions in the function app.
+        """
+        return [function_builder.build(self.auth_level) for function_builder
+                in self._function_builders]
 
-        :return: Decorator function.
+    def register_functions(self, function_container: DecoratorApi) -> None:
```
vrdmr, line 1488: @YunchuWang - As this is only used for registering blueprints, should be call is `register_blueprint`? or you are using this to register for other other generic cases? We can use pointers to functions and use specific handlers for such cases.

Define a generic _register_functions
```
def _register_functions(...):
 ...
```

and then 
```
register_blueprint=_register_functions
register_wsgiapp=_register_functions
...
```

such that it is easier to understand what the method would actually be used for.

azure/functions/decorators/function_app.py
```diff
@@ -1326,140 +1463,104 @@ def decorator():
...
         return wrap
 
-    def generic_trigger(self,
-                        arg_name: str,
-                        type: str,
-                        data_type: Optional[Union[DataType, str]] = None,
-                        **kwargs
-                        ) -> Callable:
+
+class FunctionRegister(DecoratorApi, HttpFunctionsAuthLevelMixin, ABC):
```
vrdmr, line 1467: So - this would be used by the Worker now?

/cc @gavin-aguiar

azure/functions/decorators/function_app.py
```diff
@@ -1326,140 +1463,104 @@ def decorator():
...
 
-        Ref: https://aka.ms/eventgridtrigger
 
-        :param arg_name: the variable name used in function code for the
-         parameter that receives the event data.
-        :param data_type: Defines how Functions runtime should treat the
-        parameter value.
-        :return: Decorator function.
-        """
+class BluePrint(TriggerApi, BindingApi):
```
vrdmr, line 1517: Rename to `Blueprint` - its a single word and not CamelCased.

azure/functions/decorators/function_app.py
```diff
@@ -1326,140 +1463,104 @@ def decorator():
...
-                fb.add_trigger(
-                    trigger=EventGridTrigger(
-                        name=arg_name,
-                        data_type=parse_singular_param_to_enum(data_type,
-                                                               DataType),
-                        **kwargs))
-                return fb
 
-            return decorator()
+class ThirdPartyHttpFunctionApp(FunctionRegister, TriggerApi, ABC):
```
vrdmr, line 1526: We should have a better name for this - `ExternalHttpFunctionApp`, etc.

azure/functions/decorators/function_app.py
```diff
@@ -1326,140 +1463,104 @@ def decorator():
...
-        The event_grid_trigger decorator adds
-        :class:`EventGridTrigger`
-        to the :class:`FunctionBuilder` object
-        for building :class:`Function` object used in worker function
-        indexing model. This is equivalent to defining event grid trigger
-        in the function.json which enables function to be triggered to
-        respond to an event sent to an event grid topic.
-        All optional fields will be given default value by function host when
-        they are parsed by function host.
+        super().__init__(auth_level=http_auth_level)
```
vrdmr, line 1514: we have to think for the future - if auth_level is set and no HTTP function is registered - we should give some sort of a warning - just in case. Not for this PR.

vrdmr (approved):
LGTM

azure/functions/decorators/function_app.py
YunchuWang: this is for checking one top level functionregister trying to register another top level functionregister instance which wont be allowed as functionregister object should register non-functionregister object only.

azure/functions/decorators/function_app.py
YunchuWang: will add a task for this

azure/functions/decorators/function_app.py
YunchuWang: it can be used to register custom defined classes extending triggerapi or bindingapi or both. i am thinking aliasing a register_blueprint and keeping register_functions public to register any other types of custom defined function container classes?

azure/functions/decorators/function_app.py
YunchuWang: yes, need to make change on worker to index FunctionRegister instead of functionapp

PR #118, opened 2022-04-14
Add SQL Bindings

lucyzhang929:


azure/functions/sql.py
```diff
@@ -0,0 +1,76 @@
...
+        if not isinstance(rows, list):
+            rows = [rows]
+
+        return sql.SqlRowList(
+            (None if row is None else sql.SqlRow.from_dict(row))
+            for row in rows)
+
+    @classmethod
+    def encode(cls, obj: typing.Any, *,
+               expected_type: typing.Optional[type]) -> meta.Datum:
```
YunchuWang, line 54: where is expected_type used?

azure/functions/_sql.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @classmethod
+    def from_dict(cls, dct: dict) -> 'SqlRow':
+        """Create a SqlRow from a dict object"""
+        return cls({k: v for k, v in dct.items()})
+
+    def to_json(self) -> str:
+        """Return the JSON representation of the SqlRow"""
+        return json.dumps(dict(self))
+
```
YunchuWang, line 29: can we add tests to address codecov warnings?

YunchuWang:
can we add description for this PR?

azure/functions/sql.py
lucyzhang929: Not sure what exactly this is used for but I see that all the other binding types encode function also have this. Ex. https://github.com/Azure/azure-functions-python-library/blob/3bf08e13539ba957c7df4230fc9b27fc860486fe/azure/functions/cosmosdb.py#L54

Charles-Gagnon:
@YunchuWang @vrdmr @gavin-aguiar This can be closed

PR #119, opened 2022-04-15
remove internal cosmosdb keys filtering

YunchuWang:
Change to address BUG: Cosmos DB Document should not be filtering _ts  address  https://github.com/Azure/azure-functions-python-library/issues/120

### DIFF
--- a/azure/functions/_cosmosdb.py
+++ b/azure/functions/_cosmosdb.py
@@ -7,10 +7,6 @@ import json
 from . import _abc
 
 
-# Internal properties of CosmosDB documents.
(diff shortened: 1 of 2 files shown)


gavin-aguiar:
We should add one of the removed syskeys like etag to one of the existing tests in test_cosmosdb  and check if len(doc.items) matches the data. That should take care of the code coverage tests.

vrdmr (approved):
LGTM.

minor: Please add a corresponding issue in WorkerRepo to add this check to the CosmosDB E2E test.

PR #124, opened 2022-05-03
Add SQL Binding support

Charles-Gagnon:
Adds Python language support for SQL Input/Output bindings

https://github.com/Azure/azure-functions-sql-extension

(this is a continuation of https://github.com/Azure/azure-functions-python-library/pull/118 - that one can be closed)


YunchuWang (approved):

gavin-aguiar (approved):

azure/functions/_abc.py
```diff
@@ -422,3 +422,32 @@ class OrchestrationContext(abc.ABC):
     @abc.abstractmethod
     def body(self) -> str:
         pass
+
+
+class SqlRow(abc.ABC):
```
vrdmr, line 427: @YunchuWang @gavin-aguiar - Is there a ClassWide coverage NoQA to skip the coverage the abstract class?

vrdmr (approved):
LGTM.

Please add a NoQA flag on the code lines in the abstract class which cannot be covered.

azure/functions/_abc.py
Charles-Gagnon: This should do it right?

`# flake8: noqa`

https://github.com/Azure/azure-functions-python-library/blob/dev/azure/functions/_thirdparty/werkzeug/_compat.py#L1

That seems like something that should be done separately though. Or is this a blocking issue for getting this PR in?

vrdmr: This isn't blocking but would be good to get that change in along with the code it's checking. Lets get this PR in and we can get the NoQA later.

PR #125, opened 2022-05-05
GH#107: Skip setting multi Set-Cookie headers for python 3.7

YunchuWang:
Skipping setting multiple Set-Cookie headers as SimpleCookie which is a standard python lib api in python3.7 does not support samesite attribute and would cause cookie parsing error. (Ref: https://docs.python.org/3/library/http.cookies.html). A warning will be given to customers who uses python3.7 and tries to set multiple cookie headers.

### DIFF
--- a/azure/functions/http.py
+++ b/azure/functions/http.py
@@ -2,6 +2,8 @@
 # Licensed under the MIT License.
 
 import json
+import logging
+import sys
 import typing
 from http.cookies import SimpleCookie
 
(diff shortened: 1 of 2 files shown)


gavin-aguiar (approved):

azure/functions/http.py
```diff
@@ -96,10 +98,17 @@ def encode(cls, obj: typing.Any, *,
                 datum_body = meta.Datum(type='bytes', value=b'')
 
             cookies = None
-            if "Set-Cookie" in headers:
-                cookies = [SimpleCookie(cookie) for cookie in
-                           headers.get_all('Set-Cookie')]
-                headers.pop("Set-Cookie")
+
+            if sys.version_info.major == 3 and sys.version_info.minor <= 7:
```
vrdmr, line 102: Just checking - there is a helper function - which_python types, somewhere which we've been using in many places. Could you please check?

vrdmr: Also - A comment on the why we are skipping would help.

YunchuWang: It is in the worker, checking for specific version, i prefer checking for 3.7 or below cuz 3.6 is phasing out in 09/22
image

PR #131, opened 2022-05-25
Performance tweaks for to_camel_case

pamelafox:
This PR includes two performance tweaks for the `to_camel_case` function:

1) It replaces the generator expression passed to `str.join()` with a list comprehension. The list comprehension is actually 10-15% faster as `str.join()` turns its arguments into a list anyway (if they're not already). More details in this benchmark test: https://github.com/tonybaloney/anti-patterns/pull/4

2) It removes the re.IGNORECASE flag and instead specifies a-zA-Z instead of a-z. That seems to be 20-25% faster according to benchmark in https://github.com/tonybaloney/anti-patterns/pull/5


azure/functions/decorators/utils.py
```diff
@@ -8,10 +8,8 @@
 from typing import TypeVar, Optional, Union, Iterable, Type, Callable
 
 T = TypeVar("T", bound=Enum)
-SNAKE_CASE_RE = re.compile(r'^([a-z]+\d*_[a-z\d_]*|_+[a-z\d]+[a-z\d_]*)$',
-                           re.IGNORECASE)
-WORD_RE = re.compile(r'^([a-z]+\d*)$', re.IGNORECASE)
-
+SNAKE_CASE_RE = re.compile(r'^([a-zA-Z]+\d*_[a-zA-Z\d_]*|_+[a-zA-Z\d]+[a-zA-Z\d_]*)$')
+WORD_RE = re.compile(r'^([a-zA-Z]+\d*)$')
```
pamelafox, line 12: I actually think both of these could be simplified/combined into a single shorter regular expression, `(_[a-zA-Z\d]|[a-zA-Z])\w*`, but that would be a larger change and perhaps there's a reason for wanting to separate them.

vrdmr: Thanks, @pamelafox.
@YunchuWang - We can take this experience to simplify out regexes.

vrdmr (approved):

PR #138, opened 2022-07-14
log warning for function app without http functions

YunchuWang:


azure/functions/decorators/function_app.py
```diff
@@ -1620,8 +1625,15 @@ def get_functions(self) -> List[Function]:
...
         """
-        return [function_builder.build(self.auth_level) for function_builder
-                in self._function_builders]
+        functions = [function_builder.build(self.auth_level)
+                     for function_builder in self._function_builders]
+
+        if all(not function.is_http_function() for function in functions):
+            logging.warning(
+                'Auth level for the function app is not applied to non http '
+                'function app')
```
vrdmr, line 1634: Please add a link to either https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=in-process%2Cfunctionsv2&pivots=programming-language-python#http-auth or a recommendation on what customer should do - e.g. remove auth_level.

azure/functions/decorators/function_app.py
```diff
@@ -1620,8 +1625,15 @@ def get_functions(self) -> List[Function]:
 
         :return: List of functions in the function app.
         """
-        return [function_builder.build(self.auth_level) for function_builder
-                in self._function_builders]
+        functions = [function_builder.build(self.auth_level)
+                     for function_builder in self._function_builders]
+
+        if all(not function.is_http_function() for function in functions):
```
vrdmr, line 1631: Could the trigger type be checked for all, rather than adding a flag?
or a flag at the function all level that is set whenever HTTP trigger is added (to not do a loop at all)?

YunchuWang: Dont see good way to put in function app level though i know it helps performance wise. function does not have access to functionapp and if it could, will create tight coupling as functionapp already have access to functions. 

get_functions is internally used and only once so it is not that bad. I am thinking to add a new flag in functionapp and only update the flag if no http functions found.
image

vrdmr (approved):

PR #142, opened 2022-09-09
Run startup and shutdown before handle http request

carderne:
We're using this diff to work with a [stac-fastapi](https://github.com/stac-utils/stac-fastapi) app that needs to set up and tear down a DB connection for each request.

I'm not sure how generic this API is supposed to be, but this should definitely work with any Starlette implementation (relevant `startup` code [here](https://github.com/encode/starlette/blob/bc61505faef15b673bf4bf65db4927daa354d6b8/starlette/routing.py#L616-L624).

Fully understand if the scope for this library is broader, I haven't done much due diligence.


tonybaloney:
Thanks for the PR, there's a lot of edge cases and scenarios to handle, I've submitted a PR, #187 with a full implementation

PR #148, opened 2022-11-01
No temporary loop for ASGI middleware

tonybaloney:
Fixes https://github.com/Azure/azure-functions-python-library/issues/164 

get_event_loop() is deprecated in Python 3.10, but the replacement, get_running_loop() will raise a runtime error if there is no event loop, so `asyncio.run()` is a better solution, compatible with Python 3.7+.


tonybaloney:
"compatible with Python 3.7+"

Python 3.6 is failing. Are we still supporting 3.6?

gavin-aguiar (approved):

tonybaloney:
3.6 isn't supported anymore, I could add a shim for this PR to get the 3.6 test passing, but it would add extra software debt to the project for no reason.

tonybaloney:
Suggest merging after #152

tonybaloney:
@YunchuWang this is passing now, ready to merge

PR #150, opened 2022-11-12
Moved SqlRow from _abc

gavin-aguiar:
Moved sqlrow from _abc to sql.py


vrdmr (approved):
Needs tests. Please add them. Otherwise LGTM

PR #153, opened 2022-12-02
Support ASGI middleware chains by only yielding http.disconnect when response has been fully received

tonybaloney:
Fixes a bug where `http.disconnect` would be sent even though the ASGI middleware hadn't received the whole response, which yielded empty responses.

This was shown with 2 or more pieces of FastAPI middleware because it would send them in chunks.

As per the ASGI spec, `http.disconnect` should be sent when:

```
Sent to the application if receive is called after a response has been sent or after the HTTP connection has been closed. This is mainly useful for long-polling, where you may want to trigger cleanup code if the connection closes early.
```

So the receive method has been changed to only send then when all response body has been added to the buffer. 

Fixes https://github.com/Azure/azure-functions-python-worker/issues/1104


YunchuWang (approved):
LGTM

YunchuWang (approved):

PR #161, opened 2023-01-22
Added support for http.HTTPStatus in HTTPResponse

metamoof:
It should be possible to pass values from the standard library http.HTTPStatus enum to HTTPResponse

### DIFF
--- a/azure/functions/_http.py
+++ b/azure/functions/_http.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 import collections.abc
+import http
 import io
 import json
 import types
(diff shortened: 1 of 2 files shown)


YunchuWang:
LGTM, can you fix the test failures and add unit test for it?

metamoof:
@microsoft-github-policy-service agree

azure/functions/_http.py
```diff
@@ -52,6 +53,7 @@ class HttpResponse(_abc.HttpResponse):
 
     :param int status_code:
```
metamoof, line 54: I'm not sure hot to indicate `int` or `http.HTTPStatus` here

vrdmr:
Thanks a lot, @metamoof for the PR. We'll test and get this in asap. This is definitely a useful addition.

vrdmr (approved):

gavin-aguiar (approved):

PR #171, opened 2023-02-25
Adding new to_dict method for Document

### LINKED ISSUE #170: [FeatureRequest] add to_dict to func.Document for CosmosDB
**Describe the request: A clear and concise description.**
It would be great to add a `to_dict` method to the Document class so that you can immediately pass that into a class constructor with `**` and not have to do `json.reads(doc.to_json())`

vrdmr:
Fixes #170


azure/functions/_cosmosdb.py
```diff
@@ -27,6 +27,11 @@ def to_json(self) -> str:
         """Return the JSON representation of the document."""
         return json.dumps(dict(self))
 
+    def to_dict(self) -> dict:
+        """Return the document as a dict - directly using self would also work
+        as Document is ``UserDict`` subclass and behave like dict"""
+        return dict(self)
```
gavin-aguiar, line 33: Do we need to add a test for this?

vrdmr: We don't have direct way of testing `to_json`, `to_dict` and other library methods. This is actually raising that we should have unittests for them as well.

gavin-aguiar (approved):

pdthummar (approved):

PR #187, opened 2023-07-18
[WIP] ASGI lifecycle events

tonybaloney:
This PR calls the lifecycle event for startup and shutdown for an ASGI application provided by the new V2 wrapper.


tonybaloney:
Equivalent demo with quart

```python
from quart import Quart

app = Quart(__name__)

@app.route('/')
async def hello():
    return 'hello'


@app.before_serving
async def create_db_pool():
    print("running quart startup 1")

@app.before_serving
async def use_g():
    print("Run quart startup 2")

@app.while_serving
async def lifespan():
    print("Initial startup")
    yield
    print("Shutting down quart")

@app.after_serving
async def create_db_pool():
    print("running quart shutdown 1")

```

tonybaloney:
screenshot 2023-07-25 at 13 30 30
If a startup event fails, it will capture the error in the logs, fail the request and retry on the next request. The exception is not reported to the HTTP result from the initiating request.

azure/functions/decorators/function_app.py
```diff
@@ -2774,7 +2775,13 @@ def __init__(self, app,
...
         """
         super().__init__(auth_level=http_auth_level)
-        self._add_http_app(AsgiMiddleware(app))
+        self.middleware = AsgiMiddleware(app)
+        self._add_http_app(self.middleware)
+        self.startup_task_done = False
+
+    def __del__(self):
+        if self.startup_task_done:
+            asyncio.run(self.middleware.notify_shutdown())
```
YunchuWang, line 2784: curious why for shutdown, we dont raise runtimeerror if shutdown does not succeed (similar to startup)?

YunchuWang (approved):

PR #199, opened 2023-12-20
Add Support for PEP 585 Type Hinting in Azure Functions Python Worker

YunchuWang:
This PR introduces compatibility with PEP 585 type hinting in the Azure Functions Python Worker. PEP 585, introduced in Python 3.9, allows for the use of built-in collection types (like list and dict) as generic type hints, offering a more concise and readable syntax.

addressing https://github.com/Azure/azure-functions-python-worker/issues/1034

### DIFF
--- a/azure/functions/meta.py
+++ b/azure/functions/meta.py
@@ -6,7 +6,13 @@ import collections.abc
 import datetime
 import json
 import re
+import sys
+
 from typing import Dict, Optional, Union, Tuple, Mapping, Any
+if sys.version_info >= (3, 8):
+    from typing import get_origin, get_args
+else:
+    from ._thirdparty.typing_inspect import get_origin, get_args
 
 from ._thirdparty import typing_inspect
 from ._utils import (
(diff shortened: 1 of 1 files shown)


azure/functions/meta.py
```diff
@@ -16,16 +22,28 @@
...
-    )
+    """Since python 3.9, standard collection types are supported in type hint.
+    origin is the unsubscripted version of a type (eg. list, union, etc.).
+
+    If origin is not None, then the type annotation is a builtin or part of
+    the collections class.
+    """
+    origin = get_origin(annotation)
+    if sys.version_info >= (3, 9):
+        is_iterable_anno = (origin is not None
```
gavin-aguiar, line 33: Why is `typing_inspect.is_generic_type(annotation)` not included in >3.9?

hallvictoria: `is_generic_type` will fail for the `list` type -- this is a method we defined. `get_origin` expands on our definition of `is_generic_type` but includes the built-in collection types. `get_origin` supports the generic types and returns a non-None value where `is_generic_type` would return true

PR #202, opened 2024-01-12
Naive implementation of a backwards-compatible `function_name` decorator

davidmrdavid:
This PR (https://github.com/Azure/azure-functions-python-library/pulls?q=is%3Apr+is%3Aclosed+settingsapi) seems to have made a breaking change prevent DF BluePrints from using the `function_name` decorator in versions < 1.2.8 of `azure-functions-durable`.

This PR mitigates that breaking change by creating a backwards compatibility layer that allows anything that inherits from `TriggerAPI` to continue using `function_name`. I also keep the implementation of `SettingsApi`'s `function_name` to be the same as this `TriggerAPI` backwards compat layer by unifying the implementations under a `_function_name` private method inherited from `DecoratorAPI`. The approach isn't perfect, but I think it's sufficiently clean given the circumstances.

I also added a test to validate this worked


tests/decorators/test_function_app.py
```diff
@@ -489,6 +489,31 @@ def hello(name: str):
         self.assertEqual(setting.get_settings_value("function_name"),
                          "timer_function")
 
+    def test_legacy_blueprints_with_function_name(self):
+        class LegacyBluePrint(TriggerApi, BindingApi):
+            pass
+
```
davidmrdavid, line 495: notice this doesn't inherit from SettingsApi :)

azure/functions/decorators/function_app.py
```diff
@@ -419,6 +443,20 @@ def decorator():
 
         return wrap
 
+    def function_name(self, name: str,
```
gavin-aguiar, line 446: The issue about adding it here is that now it's a trigger which it's not. Thats the reason a new api was created to add decorators which were not triggers or binding.

YunchuWang:
Hi David, wonder what error customer has been from this breaking change?

vrdmr (approved):
LGTM - thanks for the fix. :)

davidmrdavid:
@gavin-aguiar: I noticed the tests are failing, any ideas on what happened?

PR #206, opened 2024-02-12
DurableClient Converter

hallvictoria:
Created a DurableClient converter class based off of the Generic converter type so that DurableClient is no longer a generic type. Only function needed was has_implicit_output

### DIFF
--- a/azure/functions/durable_functions.py
+++ b/azure/functions/durable_functions.py
@@ -126,7 +126,7 @@ class ActivityTriggerConverter(meta.InConverter,
         return True
 
 
-# Durable Function Activity Trigger
+# Durable Functions Durable Client Bindings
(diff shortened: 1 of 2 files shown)


azure/functions/durable_functions.py
```diff
@@ -124,3 +124,12 @@ def encode(cls, obj: typing.Any, *,
     @classmethod
     def has_implicit_output(cls) -> bool:
         return True
+
+
+# Durable Function Activity Trigger
```
nytian, line 129: I think this should be Durable Functions Durable Client Bindings

nytian (approved):
Thanks for the PR! There is a typo in commits. Please update it before merge.

gavin-aguiar (approved):

PR #207, opened 2024-03-08
Add Durable Decorators directly to Python library

davidmrdavid:
This PR attempts to provide an implementation of the Durable Functions decorators (for the V2 prog model) within the Python library itself. The trick is to simply expose the right interface but defer the implementation to the Durable Functions SDK.

This PR does just that. In particular, it:
* Adds the DF decorator interfaces to the Python library. This will enable us to implement the DF templates for the V2 programming model with ease, as all DF triggers and bindings will now be exposed from the `FunctionApp` object.
* It does **not** make the durable functions SDK a "prod" dependency of the library. The customer needs to bring the durable functions SDK within their app for this to work. If they do not depend on it, we throw a friendly error asking them to bring the DF SDK.
* It makes the DF SDK part of this package's `dev`-dependencies. This allows us to write unit tests for this behavior.


davidmrdavid:
@gavin-aguiar : are we good to merge this?

PR #223, opened 2024-06-03
fix: small memory leak

hallvictoria:
Address https://github.com/Azure/azure-functions-python-worker/issues/1500. 

Changes instances where argument type is set to {} to instead have a default value of None and initialized to {} later in the method


azure/functions/decorators/function_app.py
```diff
@@ -343,6 +343,8 @@ def function_name(self, name: str,
         additional setting fields
         :return: Decorator function.
         """
+        if setting_extra_fields is None:
+            setting_extra_fields = {}
```
gavin-aguiar, line 347: Is this needed? Or does something break when settings is none?

hallvictoria: Previously we were initializing setting_extra_fields to an empty dict. This is keeping the same format so that if something is dependent on setting_extra_fields being {} then it's not broken

gavin-aguiar (approved):

PR #236, opened 2024-08-06
refactor: unique function name validation

hallvictoria:
**Main change:**
- Refactor checking for unique function names logic into higher level class.
- `FunctionBuilder` object is at the function level. `FunctionRegister` class is at the app level, so it is more suited for validation across the app as a whole.
- Now, function name `dict` is a property of the `FunctionRegister` class. There is a new method added to the `FunctionRegister` class for validating the function names. It loops through all the functions in the apps, adds the function name to the `dict`, and throws an error if there's a repeat of function names. This validation method is added as part of `get_functions` so the non-distinct function name error is thrown in the same spot as it was previously.

**Extra change:**
- removes dummy username values from kafka tests to fix security warning


azure/functions/decorators/function_app.py
```diff
@@ -3297,8 +3287,27 @@ def get_functions(self) -> List[Function]:
                 '-bindings-http-webhook-trigger?tabs=in-process'
                 '%2Cfunctionsv2&pivots=programming-language-python#http-auth')
 
+        FunctionRegister.validate_functions(self, functions=functions)
+
         return functions
 
+    def validate_functions(self, functions: List[Function]) -> bool:
```
gavin-aguiar, line 3294: Should we rename this to vaidation_function_names since we are only checking for function names?
Also it wouldn't get mixed up with the other _validation_function  method

gavin-aguiar, line 3294: We can mark this as static

azure/functions/decorators/function_app.py
```diff
@@ -3276,6 +3265,7 @@ def __init__(self, auth_level: Union[AuthLevel, str], *args, **kwargs):
         DecoratorApi.__init__(self, *args, **kwargs)
         HttpFunctionsAuthLevelMixin.__init__(self, auth_level, *args, **kwargs)
         self._require_auth_level: Optional[bool] = None
+        self.functions_bindings: Optional[Dict[Any, Any]] = None
 
     def get_functions(self) -> List[Function]:
```
gavin-aguiar, line 3270: We need to make sure that all the places where this is called in the worker have exception handling setup correctly. Otherwise, this will throw an error and without a catch, python will exit with code 1

azure/functions/decorators/function_app.py
```diff
@@ -3297,8 +3287,27 @@ def get_functions(self) -> List[Function]:
...
+            self.functions_bindings = {}
+        for function in functions:
+            function_name = function.get_function_name()
+            if self.functions_bindings.get(function_name, None):
+                raise ValueError(
+                    f"Function {function_name} does not have a unique"
+                    f" function name. Please change @app.function_name() or"
+                    f" the function method name to be unique.")
+            self.functions_bindings[function_name] = function.get_bindings()
+        return True
```
gavin-aguiar, line 3309: We don't need to return anything here

azure/functions/decorators/function_app.py
```diff
@@ -3297,8 +3287,27 @@ def get_functions(self) -> List[Function]:
...
+    def validate_functions(self, functions: List[Function]) -> bool:
+        """The functions_bindings dict contains the function name and
+        its bindings for all functions in an app. If a previous function
+        has the same name, indexing will fail here.
+        """
+        if not self.functions_bindings:
+            self.functions_bindings = {}
+        for function in functions:
+            function_name = function.get_function_name()
+            if self.functions_bindings.get(function_name, None):
```
gavin-aguiar, line 3303: We don't need to get the value of the function_name right? Can be changed to

```suggestion
            if function_name in self.functions_bindings:
```

azure/functions/decorators/function_app.py
```diff
@@ -3297,8 +3287,27 @@ def get_functions(self) -> List[Function]:
...
+                raise ValueError(
+                    f"Function {function_name} does not have a unique"
+                    f" function name. Please change @app.function_name() or"
+                    f" the function method name to be unique.")
+            self.functions_bindings[function_name] = function.get_bindings()
```
gavin-aguiar, line 3308: Why another call to get_bindings if we only need the function name?

azure/functions/decorators/function_app.py
hallvictoria: I return true for testing. It doesn't do anything per se, but it might be helpful for debugging purposes

azure/functions/decorators/function_app.py
hallvictoria: Dicts aren't iterable. If I changed it to a list that would work, but it would be N time complexity, which is why I made it a dict

azure/functions/decorators/function_app.py
hallvictoria: The value of the key-value pair in the dict right now is the corresponding function bindings. It's an O(1) call since it's just returning the corresponding variable. I thought it might be helpful for future records / debugging

azure/functions/decorators/function_app.py
hallvictoria: I don't think it can be static.

The `@staticmethod` decorator removes the `self` param, which is needed for accessing the function names dict.
I also feel it shouldn't be moved out of the `FunctionRegister` class, since it fits within the logic and the function names dict belongs to that class.

azure/functions/decorators/function_app.py
hallvictoria: The functions_bindings object is only referenced in the validate_function_names method, which has a check for None there.

I initially set the value to None to be consistent with the other properties (_require_auth_level), but I think setting it to an empty dict would also work. I'm not sure if that would have the same potential memory leak issue as addressed in https://github.com/Azure/azure-functions-python-library/pull/223 though.

gavin-aguiar: Sorry, this comment was meant for the get_functions method.

azure/functions/decorators/function_app.py
gavin-aguiar: I didn't get it. We are not changing it to a list.. We can just check if the function name is a key in bindings dictionary

azure/functions/decorators/function_app.py
gavin-aguiar: we can keep the dictionary as
{
    function_name1: 1
    function_name2: 1
}

since we are only interested in the keys the values won't matter right?
We avoid another method calling hop in that way

gavin-aguiar (approved):

PR #254, opened 2024-10-08
test: run tests for 312

hallvictoria:
- Run tests for Python 3.12
- Update README
- Update CODEOWNERS

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+import sys
+
 from setuptools import find_packages, setup
 from azure.functions import __version__
 
 EXTRA_REQUIRES = {
     'dev': [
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -17,6 +18,15 @@
     ]
 }
 
+if sys.version_info[:2] <= (3, 11):
+    EXTRA_REQUIRES.get('dev').append(
+        "flake8~=4.0.1"
+    )
+else:
+    EXTRA_REQUIRES.get('dev').append(
```
gavin-aguiar, line 26: What is this for?  Can't we directly update flake8?

gavin-aguiar (approved):

setup.py
hallvictoria: Discussed outside of GH -- the flake8 version we have pinned to works for python 37-311 but fails for 312. There currently isn't a version that works for 37 and 312, so we have to install a specific version depending on python version. We can revisit this in the future when removing 37 tests.

PR #256, opened 2024-10-09
fix: allow source to be string and enum

hallvictoria:
With the changes from https://github.com/Azure/azure-functions-python-library/pull/249, only `BlobSource` values were accepted. If the user put a string, indexing would fail.

This changes the `source` parameter to accept both `BlobSource` and string values.

### DIFF
--- a/azure/functions/decorators/blob.py
+++ b/azure/functions/decorators/blob.py
@@ -17,7 +17,10 @@ class BlobTrigger(Trigger):
                  **kwargs):
         self.path = path
         self.connection = connection
-        self.source = source.value if source else None
+        if type(source) is BlobSource:
+            self.source = source.value if source else None
+        else:
+            self.source = source
         super().__init__(name=name, data_type=data_type)
 
     @staticmethod
--- a/azure/functions/decorators/blob.py
+++ b/azure/functions/decorators/blob.py
@@ -17,8 +17,8 @@ class BlobTrigger(Trigger):
                  **kwargs):
         self.path = path
         self.connection = connection
-        if type(source) is BlobSource:
-            self.source = source.value if source else None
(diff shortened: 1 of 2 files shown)


azure/functions/decorators/blob.py
```diff
@@ -17,7 +17,10 @@ def __init__(self,
                  **kwargs):
         self.path = path
         self.connection = connection
-        self.source = source.value if source else None
+        if type(source) is BlobSource:
```
gavin-aguiar: if isinstance(source, BlobSource).

use isintance(). More robust

azure/functions/decorators/blob.py
```diff
@@ -17,7 +17,10 @@ def __init__(self,
                  **kwargs):
         self.path = path
         self.connection = connection
-        self.source = source.value if source else None
+        if type(source) is BlobSource:
+            self.source = source.value if source else None
```
gavin-aguiar: If we are checking the blobsource, it should always have a value right? Do we need another condition here?

azure/functions/decorators/blob.py
```diff
@@ -17,7 +17,10 @@ def __init__(self,
                  **kwargs):
         self.path = path
         self.connection = connection
-        self.source = source.value if source else None
+        if type(source) is BlobSource:
+            self.source = source.value if source else None
+        else:
+            self.source = source
```
gavin-aguiar: If its a string, do we check for the case?

hallvictoria: Discussed, not needed. Will be recommended for users to use the enum instead. If they use the string version, the case will be recommended to be "EventGrid"

gavin-aguiar (approved):

PR #266, opened 2024-11-18
fix: correct cosmosdb trigger types

### LINKED ISSUE #1608: [Bug] Incorrect types in cosmos_db_trigger decorator
### Expected Behavior

When using the `start_from_time` and `start_from_beginning` parameters, no typing errors should occur.

### Actual Behavior

These fields (`start_from_time` and `start_from_beginning` ) are both typed as `Optional[time]`. It seems to me like it should be:

```
start_from_beginning: Optional[bool] = None,
start_from_time: Optional[str] = None, <-- is this str or datetime? It is unclear to me
```
Also, the fields are not correctly documented in microsoft learn:
https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-cosmosdb-v2-trigger?tabs=python-v2%2Cisolated-process%2Cextensionv4%2Cnodejs-v4&pivots=programming-language-python#example
### Steps to Reproduce

Go to [this link](https://github.com/Azure/azure-functions-python-library/blob/48a5758f636abbe3f1b0634881b7eefbb74c3cf3/azure/functions/decorators/function_app.py#L1034-L1035) and you will see the incorrect types

### Relevant code being tried

```shell

```

### Relevant log output

```shell

```

### requirements.txt file

```shell

```

### Where are you facing this problem?

Local - Core Tools

### Function app name

_No response_

### Additional Information

_No response_

hallvictoria:
Fixes: https://github.com/Azure/azure-functions-python-worker/issues/1608


vrdmr (commented):
Would this break an existing app that would have taken dependency on `time`?

hallvictoria:
> Would this break an existing app that would have taken dependency on `time`?

`datetime` is one of the built-in Python modules -- not a dependency of the library. I don't think we recommend apps taking dependencies on those

PR #272, opened 2025-03-18
build: move library release to internal

hallvictoria:
Migrate library release pipeline to 1ES.

### DIFF
--- a/eng/ci/library-release.yml
+++ b/eng/ci/library-release.yml
@@ -6,7 +6,7 @@ resources:
       ref: refs/tags/release
 
 extends:
-  template: v1/1ES.Unofficial.PipelineTemplate.yml@1es
+  template: v1/1ES.Official.PipelineTemplate.yml@1es
   parameters:
     pool:
       name: 1es-pool-azfunc-public
--- a/eng/ci/library-release.yml
+++ b/eng/ci/library-release.yml
@@ -1,0 +1,26 @@
+resources:
+  repositories:
+    - repository: 1es
+      type: git
+      name: 1ESPipelineTemplates/1ESPipelineTemplates
+      ref: refs/tags/release
+
+extends:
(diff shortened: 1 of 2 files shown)


gavin-aguiar (approved):

eng/ci/library-release.yml
```diff
@@ -0,0 +1,26 @@
...
+      name: 1es-pool-azfunc-public
+      image: 1es-windows-2022
+      os: windows
+    sdl:
+      codeql:
+         compiled:
+           enabled: true # still only runs for default branch
+         runSourceLanguagesInSourceAnalysis: true
+    settings:
+      skipBuildTagsForGitHubPullRequests: ${{ variables['System.PullRequest.IsFork'] }}
```
manvkaur, line 21: this may not be required for release pipeline

build tags were being skipped for public build since we needed to run public build from fork PRs.
code ql should only run on build pipeline, are we building anything again in this pipeline?

eng/ci/library-release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - repository: 1es
+      type: git
+      name: 1ESPipelineTemplates/1ESPipelineTemplates
+      ref: refs/tags/release
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1es
+  parameters:
+    pool:
+      name: 1es-pool-azfunc-public
```
manvkaur, line 12: pipelines in internal repo using official templates should use pool - `1es-pool-azfunc` to separate internal and public workload

eng/ci/library-release.yml
hallvictoria: For CodeQL, the pipeline is not building anything, but it is downloading a build artifact from another pipeline. Can CodeQL be skipped in this case?

manvkaur: yeah, we don't need to add it. Also, official templates take care of adding itself if required.

PR #285, opened 2025-05-19
feat: allow different json library imports

EvanR-Dev:
Allow users to import orjson in their requirements.txt - the library will use that over the standard json import if installed. Here is a sample measure of performance differences. The following code calculates the average time for how long it takes to deserialize and serialize a Cosmos document JSON (~5 KB), over 10,000 runs

Note: measured in microseconds, not milliseconds

```
@app.route(route="items", auth_level=func.AuthLevel.ANONYMOUS)
@app.cosmos_db_input(arg_name="inputDocument", 
                     database_name="test",
                     container_name="items",
                     id="test-json",
                     connection="pythonworker37cdb_DOCUMENTDB")
def test_function(req: func.HttpRequest, inputDocument: func.DocumentList) -> str:
    if not inputDocument:
        return "Not found"
    
    runs = 10_000
    start = time.perf_counter()
    for _ in range(runs):
        json_string = func.Document.to_json(inputDocument[0])
        _ = func.Document.from_json(json_string)

    end = time.perf_counter()
    avg_ms = ((end - start) / runs) * 1_000_000

    logging.info(f"Time taken: {avg_ms:.2f}ms")

    return "Item fetched successfully"
```

import json
image

import orjson
image

(Not being added, but just for comparison):
import ujson
image

import simplejson
image


azure/functions/_cosmosdb.py
```diff
@@ -2,9 +2,9 @@
 # Licensed under the MIT License.
 
 import collections
-import json
 
 from . import _abc
+from ._jsonutils import json
```
EvanR-Dev, line 7: Only adding/importing here until we are ok to proceed with this implementation for all other types/files

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def dumps(self, obj, **kwargs) -> str:
+        return self.json.dumps(obj, **kwargs)
+
+    def loads(self, s: str):
+        return self.json.loads(s)
+
+
+json_impl = None
+for adapter_cls in (OrJsonAdapter, UJsonAdapter, SimpleJsonAdapter, StdJsonAdapter):
```
EvanR-Dev: There is a rare possibility that a user may pip install more than one json library in their venv - so what will be used is the order you see here with orjson as first priority if there are collisions. We could add a warning log here if we find this out - but I am seeing it's unconventional/rare to log in the library?

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,88 @@
...
+    try:
+        json_impl = adapter_cls()
+        break
+    except ImportError:
+        continue
+
+
+def dumps(obj, **kwargs) -> str:
+    if json_impl is None:
+        raise ImportError("No JSON adapter found")
```
hallvictoria: Just checking -- this should never be the case, right? The last `json_impl` checked is `StdJsonAdapter`, which is just basic `json`. If the customer does not bring any of `orjson`, `ujson`, or `simplejson`, it will default back to `json`, which is what is currently used today.

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def dumps(self, obj, **kwargs) -> str:
+        return self.json.dumps(obj, **kwargs)
+
+    def loads(self, s: str):
+        return self.json.loads(s)
+
+
+json_impl = None
+for adapter_cls in (OrJsonAdapter, UJsonAdapter, SimpleJsonAdapter,
```
hallvictoria: Which `json` is the most commonly brought or used today? We can search first for that one, then go in order with `json` being last still.

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,88 @@
...
+        return self.simplejson.loads(s)
+
+
+class StdJsonAdapter(JsonInterface):
+    def __init__(self):
+        import json
+        self.json = json
+
+    def dumps(self, obj, **kwargs) -> str:
+        return self.json.dumps(obj, **kwargs)
```
hallvictoria: Checking -- do the other jsons allow for `**kwargs` to be sent to `dumps()`?

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,88 @@
...
+import types
+
+
+class JsonInterface(ABC):
+    @abstractmethod
+    def dumps(self, obj) -> str:
+        pass
+
+    @abstractmethod
+    def loads(self, s: str):
```
hallvictoria: small nit: can there be type hints for both `dumps()` and `loads`?
Eg: `def dumps(self, obj: Typing.Any) -> str)`, `def loads(self, s: str) -> Typing.Any:`

azure/functions/_jsonutils.py
hallvictoria: +1 for adding a log. We don't have to check if the customer brings multiple json libraries, but we can log that we found either orjson, ujson, or simplejson and are using that respective one. It can be a debug log as well.

No issues with having library logs. They'll be surfaced up to the worker and displayed to the customer like worker logs

azure/functions/_jsonutils.py
EvanR-Dev: Yes, the order here is what I would keep. Orjson is the most performant and really common

azure/functions/_jsonutils.py
EvanR-Dev: They do not - only std. When replacing the rest, we can always fallback to the std for those cases where kwargs is used, but seems rare. If I dont see any usages, then I will remove this entirely

azure/functions/_jsonutils.py
EvanR-Dev: Correct - this is to satisfy the Python interpreter because it is only seeing at the time that json_impl = None. I've changed the impl to start with the std json class

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,88 @@
+from abc import ABC, abstractmethod
+import types
+
```
swapnil-nagar, line 25: NIT: Some code docs would be helpful

azure/functions/_cosmosdb.py
```diff
@@ -2,9 +2,9 @@
 # Licensed under the MIT License.
```
swapnil-nagar, line 3: Please add unit tests

azure/functions/_jsonutils.py
swapnil-nagar: Do you need messaging to provide a fix for customers?
No JSON adapter found. Please install ujson or orjson to mitigate the issue for customers

azure/functions/_jsonutils.py
EvanR-Dev: For now since we are only using orjson we can come back to this. It isn't too straightforward because the modules are loaded before the worker logging is configured, so we would have to log in dumps/loads which is not ideal. Since it's only orjson I am ok with leaving as is

azure/functions/_jsonutils.py
EvanR-Dev: This could never be the case because the default json lib is standard. So any issues would always be caught in the customer's code before the host can start correctly (for example using json.dumps but not doing an import json in their function app file)

We always just fallback to the stdlib since it does not require a pip install

azure/functions/_cosmosdb.py
EvanR-Dev: I believe there are already and pass - this wouldn't change the functionality as the result is the same but the perf is improved

gavin-aguiar:
Update the description and keep only ORJSON

azure/functions/_jsonutils.py
```diff
@@ -0,0 +1,70 @@
...
+        pass
+
+    @abstractmethod
+    def loads(self, s: Union[str, bytes, bytearray]) -> Any:
+        pass
+
+
+class OrJsonAdapter(JsonInterface):
+    def __init__(self):
+        import orjson
```
gavin-aguiar: Can we add the imports at the top?
Something like 
```
try:
    import orjson as _orjson
except ImportError:
    _orjson = None
    
 # Standard library is always present
import json as _std_json
```


Right now the code tries to create an OrJsonAdapter first, and the import of orjson happens inside that object's __init__. If orjson is missing, Python

1. Builds the adapter instance.
2. Runs its constructor, which fails at the import orjson line.
3. Raises ImportError.
4. Discards the half-built object. 
5. Enters the except block and builds a second object of type StdJsonAdapter.

Adding it at the top will be faster since the decision is made before any object construction.

gavin-aguiar (approved):

hallvictoria (approved):

EvanR-Dev:
Orjson is especially effective when the JSON payload increases in size. The improvements range from 15% (~1 KB payload) to even above 100% as the payload size increases

In these scenarios we have an EventHub trigger with an EventHub output binding that receives HTTP requests and sends it to an EventHub after deserializing the payload

Orjson:
image

Std json:
image
(some were not able to respond back in time as well)

If we purposely deserialize and serialize 10,000 times using the Cosmos library and measure the avg time (again)

Orjson:
image

Std json:
image

PR #298, opened 2025-09-12
feat: improve JSON error logging for Durable

andystaples:
Currently if to_json fails, we discard the original failure and raise our own ValueError - this PR makes the new exception inherit from the original exception, making debugging serialization issues easier.

### DIFF
--- a/azure/functions/durable_functions.py
+++ b/azure/functions/durable_functions.py
@@ -115,9 +115,9 @@ class ActivityTriggerConverter(meta.InConverter,
         try:
             callback = _durable_functions._serialize_custom_object
             result = json.dumps(obj, default=callback)
-        except TypeError:
+        except TypeError as e:
             raise ValueError(
-                f'activity trigger output must be json serializable ({obj})')
+                f'activity trigger output must be json serializable ({obj})') from e
 
         return meta.Datum(type='json', value=result)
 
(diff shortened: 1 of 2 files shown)


azure/functions/durable_functions.py
```diff
@@ -99,10 +99,10 @@ def decode(cls,
             except json.JSONDecodeError:
                 # String failover if the content is not json serializable
                 result = data.value
-            except Exception:
+            except Exception as e:
                 raise ValueError(
                     'activity trigger input must be a string or a '
-                    f'valid json serializable ({data.value})')
+                    f'valid json serializable ({data.value})') from e
```
hallvictoria, line 105: nit: could you add a unit test for this in `tests/test_durable_functions.py`?

andystaples: Done - let me know if you need to see more cases - I figured as long as the error was there in one case, that was sufficient

hallvictoria: LGTM - this change can be part of v1.25.0. We're looking to release 1.25.0b1 by end of October - lmk if there's any concerns over timelines!

andystaples: Sounds good to me

hallvictoria (approved):

gavin-aguiar (approved):

PR #304, opened 2025-11-05
fix: Handle comma-separated values in x-forwarded-port header

lucadelauw:
When Azure Functions applications are deployed behind proxies or load balancers (such as Azure Application Gateway), the 'x-forwarded-port' header may contain multiple comma-separated port values (e.g., "443,8080,433"). The current implementation attempts to convert the entire string to an integer, which causes an error and subsequent function crash.

My fix extracts the first port value from this header and also strips whitespaces. Also included new tests for these

see: Azure/azure-functions-python-worker#1768

### DIFF
--- a/azure/functions/_http_wsgi.py
+++ b/azure/functions/_http_wsgi.py
@@ -113,7 +113,9 @@ class WsgiRequest:
     def _get_port(self, parsed_url, lowercased_headers: Dict[str, str]) -> int:
         port: int = 80
         if lowercased_headers.get('x-forwarded-port'):
-            return int(lowercased_headers['x-forwarded-port'])
+            # Handle comma-separated port values (e.g., "443,8080,433")
+            forwarded_port = lowercased_headers['x-forwarded-port'].split(',')[0].strip()
+            return int(forwarded_port)
         elif getattr(parsed_url, 'port', None):
             return int(parsed_url.port)
         elif parsed_url.scheme == 'https':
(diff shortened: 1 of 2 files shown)


lucadelauw:
@microsoft-github-policy-service agree

lucadelauw:
Can this please be looked at since it's quite a blocking issue. This is a fairly straightforward fix
@hallvictoria Would be highly appreciated if we can slip this into the next release. Thanks!

hallvictoria:
Hi @lucadelauw, thanks for this. Our next beta release (1.25.0b2) has a lot of changes currently, but we can look at taking this in the beta release after (1.25.0b3). ETA for b3 release will be ~end of this week to unblock this scenario

hallvictoria:
Hey @lucadelauw, I noticed that this PR doesn't fix the original scenario based on the bug report, where the `x-forwarded-port` value being passed in is `90344081.test.com`. I had a couple other comments, so thought it'd be easier if I opened a separate PR for this https://github.com/Azure/azure-functions-python-library/pull/307. Please feel free to take a look and make sure it helps your scenario!

lucadelauw:
I see. I indeed just implemented the scenario that I specifically had issues with. Your new PR seems to cover that and more. Thanks!

hallvictoria:
closing in favor of https://github.com/Azure/azure-functions-python-library/pull/307. thanks!

PR #309, opened 2025-12-05
feat: support durabletask-azurefunctions

hallvictoria:
Adding support for durabletask-azurefunctions.

1. Identify which package is being used
2. Based on package, we call the appropriate decorators
3. Based on package, we assign the appropriate converter

### DIFF
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -12,6 +12,7 @@ from .decorators import (FunctionApp, Function, Blueprint,
                          AsgiFunctionApp, WsgiFunctionApp,
                          ExternalHttpFunctionApp, BlobSource, McpPropertyType)
 from ._durable_functions import OrchestrationContext, EntityContext
+from .durable_functions import register_durable_converters
 from .decorators.function_app import (FunctionRegister, TriggerApi,
                                       BindingApi, SettingsApi)
 from .extension import (ExtensionMeta, FunctionExtensionException,
@@ -41,6 +42,9 @@ from . import durable_functions  # NoQA
 from . import sql  # NoQA
 from . import warmup  # NoQA
 from . import mysql  # NoQA
+
+
+register_durable_converters()
 
 
 __all__ = (
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -110,4 +110,4 @@ __all__ = (
     'McpPropertyType'
 )
 
-__version__ = '1.25.0b2'
+__version__ = '1.25.0b3dev1'
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -110,4 +110,4 @@ __all__ = (
     'McpPropertyType'
 )
 
-__version__ = '1.25.0b3.dev1'
+__version__ = '1.25.0b3.dev2'
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -110,4 +110,4 @@ __all__ = (
     'McpPropertyType'
 )
 
-__version__ = '1.25.0b3dev1'
+__version__ = '1.25.0b3.dev1'
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -110,4 +110,4 @@ __all__ = (
     'McpPropertyType'
 )
 
-__version__ = '1.25.0b3.dev2'
+__version__ = '1.25.0b3.dev3'
--- a/azure/functions/decorators/durable_functions.py
+++ b/azure/functions/decorators/durable_functions.py
@@ -17,7 +17,7 @@ def get_durable_package():
     If both the legacy and current packages are installed,
     we log a warning and prefer the current package.
 
-    If neither package is installed, we raise an exception.    
+    If neither package is installed, we return None.    
     """
     _logger.info("Attempting to import Durable Functions package.")
     using_legacy = False
@@ -50,12 +50,7 @@ def get_durable_package():
         "The `durabletask-azurefunctions` package will be used.")
     
     if not using_durable_task and not using_legacy:
-        error_message = \
-                "Attempted to use a Durable Functions decorator, " \
-                "but the `durabletask-azurefunctions` SDK package could not be " \
-                "found. Please install `durabletask-azurefunctions` to use " \
-                "Durable Functions."
-        raise Exception(error_message)
+        return None
 
     df = durable_functions
 
--- a/azure/functions/decorators/durable_functions.py
+++ b/azure/functions/decorators/durable_functions.py
@@ -20,30 +20,26 @@ def get_durable_package():
     If neither package is installed, we return None.
     """
     _logger.info("Attempting to import Durable Functions package.")
-    using_legacy = False
-    using_durable_task = False
     global df
     if df:
         _logger.info("Durable Functions package already loaded. DF: %s", df)
         return df
 
+    # Check for azure-functions-durable package
     try:
-        import durabletask  # noqa
-        using_durable_task = True
-        _logger.info("`durabletask` package found.")
-    except ImportError:
-        _logger.info("`durabletask` package not found.")
-        pass
-    try:
-        import azure.durable_functions  as durable_functions # noqa
+        import azure.durable_functions as durable_functions # noqa
         _logger.info("`azure.durable_functions` package found.")
     except ImportError:
         _logger.info("`azure.durable_functions` package not found.")
-        pass
-
-    if not using_durable_task and not using_legacy:
         return None
+
+    # Check version of azure-functions-durable package
+    try:
+        if hasattr(durable_functions, 'version') and durable_functions.version.startswith("2."):
+            _logger.info("Using `azure.durable_functions` v2.x package.")
+    except Exception:
+        _logger.info("Using `azure.durable_functions` v1.x package.")
 
     df = durable_functions
 
-    return durable_functions
+    return df
--- a/azure/functions/decorators/durable_functions.py
+++ b/azure/functions/decorators/durable_functions.py
@@ -28,28 +28,18 @@ def get_durable_package():
         return df
 
     try:
-        import azure.durable_functions as durable_functions
-        using_legacy = True
-        _logger.warning("`durabletask-azurefunctions` is available now! "
-                        "See <AKA.MS LINK HERE> for more details.")
+        import durabletask  # noqa
+        using_durable_task = True
+        _logger.info("`durabletask` package found.")
     except ImportError:
-        _logger.info("`azure-functions-durable` package not found.")
+        _logger.info("`durabletask` package not found.")
         pass
     try:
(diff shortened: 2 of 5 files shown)


azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,62 @@
...
+    if df:
+        _logger.info("Durable Functions package already loaded. DF: %s", df)
+        return df
+
+    try:
+        import azure.durable_functions as durable_functions
+        using_legacy = True
+        _logger.warning("`azure-functions-durable` is deprecated. " \
+        "Please migrate to the new `durabletask-azurefunctions` package. " \
+        "See <AKA.MS LINK HERE> for more details.")
```
hallvictoria: Log a warning if `azure-functions-durable` is being used.

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        import durabletask.azurefunctions as durable_functions
+        using_durable_task = True
+    except ImportError:
+        _logger.info("`durabletask-azurefunctions` package not found.")
+        pass
+
+    if using_durable_task and using_legacy:
+        # Both packages are installed; prefer `durabletask-azurefunctions`.
+        _logger.warning("Both `azure-functions-durable` and " \
```
hallvictoria: In the case that the customer has both durable packages, log a warning & prioritize `durabletask-azurefunctions`.

azure/functions/decorators/function_app.py
```diff
@@ -347,17 +349,11 @@ def _get_durable_blueprint(self):
...
-        except ImportError:
-            error_message = \
-                "Attempted to use a Durable Functions decorator, " \
-                "but the `azure-functions-durable` SDK package could not be " \
-                "found. Please install `azure-functions-durable` to use " \
-                "Durable Functions."
-            raise Exception(error_message)
+        _logger.info("Getting Durable Functions blueprint.")
+        df = get_durable_package()
+        df_bp = df.Blueprint()
```
hallvictoria: `df` here will now be either `azure-functions-durable` or `durabletask-azurefunctions`

azure/functions/durable_functions.py
```diff
@@ -5,13 +5,17 @@
...
+_logger = logging.getLogger('azure.functions.AsgiMiddleware')
 
-# Durable Function Orchestration Trigger
-class OrchestrationTriggerConverter(meta.InConverter,
+# ---------------- Legacy Durable Functions Converters ---------------- #
+# Legacy Durable Function Orchestration Trigger
+class LegacyOrchestrationTriggerConverter(meta.InConverter,
                                     meta.OutConverter,
-                                    binding='orchestrationTrigger',
+                                    binding=None,
```
hallvictoria: Previous converters are renamed to "Legacy" with no changes. `binding` is set to `None` so that they are not automatically registered

azure/functions/durable_functions.py
```diff
@@ -39,9 +43,196 @@ def has_implicit_output(cls) -> bool:
...
+                'unexpected type of data received for the "generic" binding ',
+                repr(data_type)
+            )
+
+        return result
+
+
+# ---------------- Durable Task Durable Functions Converters ---------------- #
+# Durable Function Orchestration Trigger
+class OrchestrationTriggerConverter(meta.InConverter,
```
hallvictoria, line 203: Currently, converters for durable task package are named `OrchestrationTrigger Converter`, `EntityTriggerConverter`, `ActivityTriggerConverter`, and `DurableClientConverter` - open to renaming if preferable

azure/functions/durable_functions.py
```diff
@@ -39,9 +43,196 @@ def has_implicit_output(cls) -> bool:
...
+    def decode(cls,
+               data: meta.Datum, *,
+               trigger_metadata) -> _durable_functions.OrchestrationContext:
+        return _durable_functions.OrchestrationContext(data.value)
+
+    @classmethod
+    def encode(cls, obj: typing.Any, *,
+               expected_type: typing.Optional[type]) -> meta.Datum:
+        # Durable function context should be a string
+        return meta.Datum(type='string', value=obj)
```
hallvictoria, line 226: New converters for durable task are currently identical to legacy converters except for this change based on https://github.com/microsoft/durabletask-python/pull/75#discussion_r2578565402.

Will update these converters as needed to be compatible with durable task

azure/functions/durable_functions.py
```diff
@@ -190,3 +381,34 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
...
+    _logger.info("Registering Durable Functions converters based on ")
+    pkg = get_durable_package()
+    if pkg is None:
+        # Durable library not installed → do nothing
+        return
+
+    _logger.info("Durable Functions package loaded: %s", pkg.__name__)
+    _logger.info("Current bindings before registration: %s", meta._ConverterMeta._bindings)
+    # Clear existing bindings if they exist
+    meta._ConverterMeta._bindings.pop("orchestrationTrigger", None)
```
hallvictoria, line 385: This should never happen, but just in case - if there are bindings already registered for durable, remove them. Then we add the appropriate converter (Legacy vs. new) based on the durable package brought by the customer

azure/functions/decorators/function_app.py
```diff
@@ -347,18 +349,20 @@ def _get_durable_blueprint(self):
...
+        if df:
             df_bp = df.Blueprint()
             return df_bp
-        except ImportError:
+        else:
             error_message = \
                 "Attempted to use a Durable Functions decorator, " \
-                "but the `azure-functions-durable` SDK package could not be " \
-                "found. Please install `azure-functions-durable` to use " \
+                "but the `durabletask-azurefunctions` SDK package could not be " \
```
hallvictoria: This is the case where the customer tries to use a DF decorator (`func.FunctionApp`) but doesn't have either package installed. `df` is None, so we raise an error

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,57 @@
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License.
+import logging
+
+_logger = logging.getLogger('azure.functions.AsgiMiddleware')
```
hallvictoria: All of the info logs will be removed in the future, just for debugging purposes now

gavin-aguiar: Why AsgiMiddleware is used in durable for logging?

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,57 @@
...
+    If the `azure-functions-durable` package is installed, we
+    log a warning that this legacy package
+    is deprecated.
+
+    If both the legacy and current packages are installed,
+    we log a warning and prefer the current package.
+
+    If neither package is installed, we return None.    
+    """
+    _logger.info("Attempting to import Durable Functions package.")
```
gavin-aguiar, line 22: Should we make this log debug?

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,57 @@
...
+        return df
+
+    try:
+        import azure.durable_functions as durable_functions
+        using_legacy = True
+        _logger.warning("`azure-functions-durable` is deprecated. " \
+        "Please migrate to the new `durabletask-azurefunctions` package. " \
+        "See <AKA.MS LINK HERE> for more details.")
+    except ImportError:
+        _logger.info("`azure-functions-durable` package not found.")
```
gavin-aguiar: This can be a debug too

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,57 @@
...
+        pass
+    try:
+        import durabletask.azurefunctions as durable_functions
```
gavin-aguiar: Why do we need to check for both? If durabletask is added, lets directly use it and just add a log that durable task is being used.

azure/functions/decorators/function_app.py
```diff
@@ -347,18 +349,20 @@ def _get_durable_blueprint(self):
...
+                "found. Please install `durabletask-azurefunctions` to use " \
                 "Durable Functions."
             raise Exception(error_message)
 
+
```
gavin-aguiar: NIT; extra line added

azure/functions/__init__.py
```diff
@@ -43,6 +44,9 @@
 from . import mysql  # NoQA
 
 
+register_durable_converters()
```
gavin-aguiar, line 47: This goes inside the __all__ block

azure/functions/decorators/durable_functions.py
hallvictoria: We still need to maintain backwards compatibility with `azure-functions-durable`. We have to check for both to determine which converter / decorator is being used. This check is only done once though - after that, the `df` variable will save which module is being used

azure/functions/decorators/durable_functions.py
hallvictoria: Temporary, will be updated later

azure/functions/decorators/durable_functions.py
hallvictoria: These logs are just for debugging, will be removed / moved to debug later

azure/functions/decorators/durable_functions.py
hallvictoria: These logs are just for debugging, will be removed / moved to debug later

azure/functions/__init__.py
hallvictoria: `register_durable_converters` is not exposed to the worker. It's called at init, whenever `azure-functions` is imported by the worker.

azure/functions/decorators/durable_functions.py
andystaples: This message is probably too aggressive before we post the Azure deprecation notice, would suggest for now an info message informing about the new package only.

azure/functions/decorators/durable_functions.py
andystaples: There will be more unexpected behavior from the Durable extension if both modules are present, this warning suggests that we will safely default to the new module but that might not always be the case. Recommend changing this message to. 
"Both `azure-functions-durable` and `durabletask-azurefunctions` packages are installed. This may lead to unexpected behavior, please resolve this conflict by removing one of these packages from the Python environment. Decorators from `durabletask-azurefunctions` will be used."

azure/functions/durable_functions.py
```diff
@@ -140,6 +331,11 @@ def has_trigger_support(cls) -> bool:
 
     @classmethod
     def check_input_type_annotation(cls, pytype: type) -> bool:
+        try:
+            import azure.durable_functions as adf
```
hallvictoria: TODO: is this try-catch needed?

PR #318, opened 2026-03-17
feat: add support for structured content for MCP

### LINKED ISSUE #1825: [MCP Tool] Implement support for structured content
Implement support for MCP structured content in tool functions. This feature allows tools to return structured data that clients can programmatically consume, while maintaining backwards compatibility by also including text content.

#### Background

The MCP specification supports structured content that can be returned alongside unstructured (text) content from tool invocations. When a tool returns structured content, it should ALSO include a text representation for backwards compatibility with clients that don't support structured content.

**MCP Specification:** https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

**Reference Implementation:** https://github.com/Azure/azure-functions-mcp-extension/pull/172

#### Host Extension Contract

The host extension has a feature that allows the worker to be responsible for result content by setting `UseResultSchema` to `true` in the tool trigger binding. This was originally added to allow workers to support rich binding types such as `TextContentBlock`, `ImageContentBlock`, or `CallToolResult`. This same feature can be leveraged to enable structured content support.

```csharp
    /// <summary>
    /// Gets or sets whether the result schema should be provided by the worker instead of being generated by the host extension.
    /// When set to true, the <see cref="ResultSchema"/> property must be provided.
    /// When set to false (default), the extension generates the schema from the function return type.
    /// </summary>
    public bool UseResultSchema { get; set; } = false;
```

When `UseResultSchema` is set to `true`, the host extension expects tool results to be wrapped in an `McpToolResult` object. This is a wrapper that the worker sends back to the host containing the serialized tool result along with metadata about the result type.

> **Note:** If `UseResultSchema` is `false` (the default), the host extension handles result processing and this wrapper is not needed. The `McpToolResult` wrapper is only required when the worker wants to control the result schema.

**McpToolResult schema:**
```csharp
public class McpToolResult
{
    // The type of content (e.g., "text", "image", "audio", "multi_content_result", "call_tool_result")
    public required string Type { get; set; }
    
    // The serialized content block(s)
    public string? Content { get; set; }
    
    // NEW: Optional structured content as a JSON object
    public string? StructuredContent { get; set; }
}

When using `McpToolResult`, the worker can use this to populate the `structuredContent` field when the user wants to return structured data.

**Standard result with structured content:**
```json
{
  "type": "text",
  "content": "<serialized-TextContentBlock>",
  "structuredContent": "<serialized-structured-data>"
}
```

**CallToolResult with structured content:**
```json
{
  "type": "call_tool_result",
  "content": "<serialized-CallToolResult>",
  "structuredContent": "<serialized-structured-data>"
}
```

#### Key Considerations

1. **Backwards compatibility:** When `structuredContent` is populated, there must also be a `TextContentBlock` in the content for clients that don't support structured content.

2. **The `structuredContent` field** should contain the JSON-serialized structured data that clients can programmatically consume.

3. **The `content` field** contains the serialized content block(s) as before - this is unchanged from existing behavior.

hallvictoria:
fixes: https://github.com/Azure/azure-functions-python-worker/issues/1825, https://github.com/Azure/azure-functions-python-worker/issues/1836, https://github.com/Azure-Samples/remote-mcp-functions-python/issues/38

image

Examples:

```
# ============================================================================
# Scenario 1: Single ResourceLinkBlock return
# Matches: GetFunctionsLogo in .NET
# ============================================================================
@app.mcp_tool()
def get_functions_logo() -> func.ResourceLinkBlock:
    """Returns the Azure Functions logo as a resource link."""
    return func.ResourceLinkBlock(
        uri="file://logo.png",
        name="Azure Functions Logo",
        mime_type="image/png"
    )

# ============================================================================
# Scenario 2: Multiple content blocks (List[ContentBlock])
# Matches: MultiContentTypeFunction in .NET
# ============================================================================
@app.mcp_tool()
def multi_content_type_function(
    data: str,
    mime_type: Optional[str] = None
) -> List[func.ContentBlock]:
    """Responds to user with multiple content blocks."""
    return [
        func.TextContentBlock(text="Here is an image for you!"),
        func.ResourceLinkBlock(
            name="example",
            uri="https://www.google.com/",
            description="Image Information"
        ),
        func.ImageContentBlock(
            data=data,
            mime_type=mime_type or "image/jpeg"
        )
    ]


# ============================================================================
# Scenario 3: Single ImageContentBlock return
# Matches: RenderImage in .NET
# ============================================================================
@app.mcp_tool()
def render_image(
    data: str,
    mime_type: Optional[str] = None
) -> func.ImageContentBlock:
    """Responds to user with an image."""
    return func.ImageContentBlock(
        data=data,
        mime_type=mime_type or "image/jpeg"
    )

# ============================================================================
# Scenario 4: @mcp_content decorated dataclass
# Matches: GetSnippet returning [McpContent] Snippet in .NET
# ============================================================================
@func.mcp_content
class Snippet:
    """Snippet class that will be serialized as structured content."""
    def __init__(self, name: str, content: Optional[str] = None):
        self.name = name
        self.content = content

@app.mcp_tool()
def get_snippet(name: str) -> Optional[Snippet]:
    """Gets a code snippet by name."""
    if name in snippets_cache:
        return Snippet(name=name, content=snippets_cache[name])
    return None


# ============================================================================
# Scenario 5: CallToolResult with explicit content and structured content
# Matches: GetImageWithMetadata in .NET
# ============================================================================
@app.mcp_tool()
def get_image_with_metadata() -> func.CallToolResult:
    """Returns an image with metadata as structured content."""
    # Metadata object
    metadata = {
        "ImageId": "logo",
        "Format": "png",
        "CreatedAt": datetime.utcnow().isoformat(),
        "Tags": ["functions", "azure", "serverless"]
    }
    
    # Read image file (or use a tiny test image)
    # For demo, using a 10x10 red pixel PNG (larger version of the 1x1)
    import zlib
    import struct
    
    # PNG signature
    png_data = b'\x89PNG\r\n\x1a\n'
    
    def make_chunk(chunk_type, data):
        """Create a PNG chunk with proper CRC."""
        chunk_data = chunk_type + data
        crc = zlib.crc32(chunk_data) & 0xffffffff
        return struct.pack('>I', len(data)) + chunk_data + struct.pack('>I', crc)
    
    # IHDR: 10x10, 8-bit RGB
    width, height = 10, 10
    ihdr_data = struct.pack('>IIBBBBB', width, height, 8, 2, 0, 0, 0)
    png_data += make_chunk(b'IHDR', ihdr_data)
    
    # IDAT: image data (10x10 red pixels)
    raw_data = b''
    for y in range(height):
        raw_data += b'\x00'  # Filter type: None
        for x in range(width):
            raw_data += b'\xff\x00\x00'  # RGB: Red
    
    compressed = zlib.compress(raw_data, 9)
    png_data += make_chunk(b'IDAT', compressed)
    
    # IEND
    png_data += make_chunk(b'IEND', b'')
    
    base64_image = base64.b64encode(png_data).decode('utf-8')
    
    # Construct CallToolResult with both content blocks and structured content
    return func.CallToolResult(
        content=[
            func.TextContentBlock(text=json.dumps(metadata)),  # JSON representation
            func.ImageContentBlock(
                data=base64_image,
                mime_type="image/png"
            )
        ],
        structured_content=metadata  # Structured data for parsing
    )


# ============================================================================
# Additional scenario: use_result_schema parameter
# ============================================================================
@app.mcp_tool(use_result_schema=True)
def get_structured_data(query: str) -> Snippet:
    """Returns structured data using custom class."""
    return MyCustomClass(
        name="result",
        content=f"Query result for: {query}"
    )

```


azure/functions/decorators/function_app.py
```diff
@@ -1615,12 +1615,74 @@ def mcp_tool(self, metadata: Optional[str] = None):
...
+
+                    # Check for Optional[T] where T is an MCP type
+                    if origin is Union:
+                        for arg in args:
+                            if isinstance(arg, type(None)):
+                                continue
+                            try:
+                                if isinstance(arg, type):
+                                    if issubclass(arg, (ContentBlock, CallToolResult)):
+                                        is_content_block = True
```
gavin-aguiar, line 1674: Here if arg is CallToolResult, you're setting is_content_block = True instead of is_call_tool_result = True..
Lets break this into 2 if conditions
if issubclass(arg, ContentBlock):
...
elif issubclass(arg, CallToolResult):
...

azure/functions/decorators/function_app.py
```diff
@@ -1667,6 +1734,66 @@ async def wrapper(context: str, *args, **kwargs):
                 result = target_func(**call_kwargs)
                 if asyncio.iscoroutine(result):
                     result = await result
+
+                if result is None:
+                    return str(result)
```
gavin-aguiar: This will return "None" .  Maybe return ""

azure/functions/decorators/function_app.py
```diff
@@ -1615,12 +1615,74 @@ def mcp_tool(self, metadata: Optional[str] = None):
...
         def decorator(fb: FunctionBuilder) -> FunctionBuilder:
             target_func = fb._function.get_user_function()
             sig = inspect.signature(target_func)
 
+            # Auto-detect MCP return types and set use_result_schema=True
+            # Use a separate variable to avoid UnboundLocalError
+            auto_use_result_schema = use_result_schema
+            return_annotation = sig.return_annotation
+            if return_annotation != inspect.Signature.empty and not auto_use_result_schema:
+                from azure.functions.mcp import ContentBlock, CallToolResult
```
gavin-aguiar: These imports will happen at every call.  Can we move this to the top?

azure/functions/decorators/mcp.py
```diff
@@ -156,3 +161,70 @@ def build_property_metadata(sig,
...
+    - The object is not a primitive type (str, int, float, bool, None)
+    - The object is not a dict or list (unless explicitly marked)
+
+    This mimics the .NET implementation's McpContentAttribute checking.
+    """
+    if obj is None:
+        return False
+
+    # Primitive types don't generate structured content unless explicitly marked
+    if isinstance(obj, (str, int, float, bool)):
```
gavin-aguiar: NIT: bool is a subclass of int so its just a redundant check. (str, int, float) should be enough

azure/functions/decorators/mcp.py
```diff
@@ -156,3 +161,70 @@ def build_property_metadata(sig,
...
+
+
+def should_create_structured_content(obj: typing.Any) -> bool:
+    """
+    Determines whether structured content should be created for the given object.
+
+    Returns True if:
+    - The object's class is decorated with a marker that sets __mcp_content__ = True
+    - The object is not a primitive type (str, int, float, bool, None)
+    - The object is not a dict or list (unless explicitly marked)
```
gavin-aguiar: Should we add a check for this ? 
if isinstance(obj, (dict, list)):
    return False

azure/functions/decorators/function_app.py
hallvictoria: Done

azure/functions/decorators/mcp.py
hallvictoria: For bools, the check should short-circuit & return when it hits `int`. I lean towards keeping the check for bool just to be explicit in which primitive types are unsupported.

gavin-aguiar (approved):

PR #336, opened 2026-04-27
feat: Add registered-type JSON pipeline for Durable Functions custom objects

andystaples:
**Summary**

Refactors `azure.functions._durable_functions` to introduce an explicit registry for round-tripping user-defined classes through JSON, and reworks the existing `_serialize_custom_object` / `_deserialize_custom_object` helpers so that decoding is a pure data transformation (no module loading, no I/O).

**What changed**

- New public API in `azure.functions._durable_functions`:
  - `register_durable_serializable_type(cls)` — opt-in registration (usable as a decorator) for classes that expose `to_json` / `from_json`.
  - `to_json_string(obj)` / `from_json_string(s, *, accept_legacy=False)` — symmetric encode/decode pipeline. Plain JSON values round-trip unchanged; dicts whose shape would collide with an internal marker are escaped on encode and restored on decode; only registered classes are reconstructed from `__azfunc_obj__` markers.
- Reworked back-compat shims:
  - `_serialize_custom_object` still emits the legacy `__class__/__module__/__data__` shape, but now requires the class to be registered.
  - `_deserialize_custom_object` resolves target classes through the registry first, falling back to a hardened lookup in already-loaded modules (`sys.modules` only). The fallback requires a real `type` in the named module that defines `to_json` and a `classmethod`/`staticmethod` `from_json`. Unrecognised markers pass through as plain dicts.
- New `AZURE_FUNCTIONS_DURABLE_STRICT_LEGACY_DESERIALIZE` env flag disables the `sys.modules` fallback for callers who want registry-only resolution.
- Legacy decode emits a `DeprecationWarning` (resolved or not) so callers can migrate to the registry; the legacy shape will be removed in the next major.
- `OrchestrationContext` and `EntityContext` are unchanged.

**Tests**

`tests/test_durable_functions.py` gains coverage for:
- registry semantics (validation, idempotency, conflict detection),
- symmetric round-trip over a corpus of plain JSON values, dicts containing reserved-key shapes, and nested collisions,
- registered-instance round-trip (top-level and nested),
- a `_NoLazyImports` context manager that asserts decode never calls `importlib.import_module` across every entry point and input shape,
- the hardened resolver's accept/reject criteria (loaded vs unloaded module, instance-method `from_json`, missing `to_json`, non-class attribute, re-export with mismatched `__module__`),
- legacy-shape decode behaviour with `accept_legacy=True`, including registry precedence, `sys.modules` fallback, strict mode, unloaded module, and the `accept_legacy=False` default,
- legacy shim serialize behaviour and an end-to-end shim round-trip,
- concurrent registration safety,
- module surface (no top-level `import_module` binding).

**Compatibility**

- No wire-format change. Existing producers continue to emit the legacy shape; existing in-flight history decodes through the back-compat shim or `from_json_string(..., accept_legacy=True)`.
- Apps that defined custom classes in modules imported at startup continue to round-trip without changes; a `DeprecationWarning` invites them to call `register_durable_serializable_type(cls)`.
- Apps relying on a class whose module had not yet been imported in the worker will see those values come back as `dict` (with a warning that explains how to opt in via the registry or by importing at startup).

### DIFF
--- a/azure/functions/_durable_functions.py
+++ b/azure/functions/_durable_functions.py
@@ -1,86 +1,281 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+"""JSON serialization helpers for Durable Functions custom objects.
 
-from typing import Union
+This module exposes a registry-based, symmetric encode/decode pipeline
+for round-tripping user-defined classes through JSON, plus back-compat
+shims for the historical ``default=`` / ``object_hook=`` helper pair.
+
+The new pipeline (``to_json_string`` / ``from_json_string``) only
+reconstructs classes that have been registered via
+``register_durable_serializable_type``; dicts whose shape would collide
+with a marker are escaped on encode and restored on decode. The legacy
+shims (``_serialize_custom_object`` / ``_deserialize_custom_object``)
+preserve the original wire shape for back-compat but resolve target
+classes only through the registry or already-loaded modules; no module
+is imported during decoding.
+"""
+from __future__ import annotations
+
+import json
+import os
+import sys
+import warnings
+from threading import Lock
+from typing import Any, Dict, Type, Union
+
 from . import _abc
-from importlib import import_module
 
 
-# Utilities
-def _serialize_custom_object(obj):
-    """Serialize a user-defined object to JSON.
+_OBJ_KEY = "__azfunc_obj__"
+_ESC_KEY = "__azfunc_escaped__"
+_LEGACY_KEYS = ("__class__", "__module__", "__data__")
 
-    This function gets called when `json.dumps` cannot serialize
-    an object and returns a serializable dictionary containing enough
-    metadata to recontrust the original object.
+_registered_types: Dict[str, Type] = {}
+_registry_lock = Lock()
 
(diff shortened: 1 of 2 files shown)


azure/functions/_durable_functions.py
```diff
@@ -1,86 +1,281 @@
...
+            f"Durable Functions saw a legacy custom-object marker for "
+            f"{mod_name}.{cls_name} but could not resolve it. "
+            "Returning as dict. Either register the class with "
+            "register_durable_serializable_type(cls), or import the "
+            "module at app startup and ensure the class defines "
+            "to_json/from_json (from_json must be a classmethod or "
+            "staticmethod). The legacy shape will be removed in the next "
+            "major version."
+        )
+    warnings.warn(msg, DeprecationWarning, stacklevel=3)
```
hallvictoria, line 171: [checking] does this warning show up? We may need to use logging

```
import logging
_logger = logging.getLogger('azure.functions.DurableFunctions')
```

andystaples: Addressed - also logs using `logging` with the suggested durable logger

PR #353, opened 2026-07-22
feat: Add support for new Durable Python SDK

andystaples:


azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,43 @@
...
+    If the `azure-functions-durable` package is installed, we
+    log a warning that this legacy package
+    is deprecated.
+
+    If both the legacy and current packages are installed,
+    we log a warning and prefer the current package.
+
+    If neither package is installed, we return None.
+    """
+    _logger.info("Attempting to import Durable Functions package.")
```
hallvictoria: Please make all logs except for essential ones `debug`

azure/functions/__init__.py
```diff
@@ -116,4 +120,4 @@
     'mcp_content',
 )
 
-__version__ = '2.3.0b1'
+__version__ = '2.3.0b2'
```
hallvictoria: Please change back to `b1` - the version bump is part of the release process

azure/functions/durable_functions.py
```diff
@@ -199,3 +205,207 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
...
+    """
+    Registers the appropriate Durable Functions converters based on the
+    installed Durable Functions package.
+    """
+    _logger.info("Registering Durable Functions converters based on ")
+    pkg = get_durable_package()
+    if pkg is None:
+        return
+
+    _logger.info("Durable Functions package loaded: %s", pkg.__name__)
```
hallvictoria: These logs can probably be removed completely, or at least moved to debug if you want to keep them for debugging later on

azure/functions/durable_functions.py
```diff
@@ -199,3 +205,207 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
...
+    meta._ConverterMeta._bindings.pop("activityTrigger", None)
+    meta._ConverterMeta._bindings.pop("durableClient", None)
+
+    if not hasattr(pkg, 'version'):
+        _logger.info("Registering Legacy Durable Functions converters.")
+        meta._ConverterMeta._bindings["orchestrationTrigger"] = LegacyOrchestrationTriggerConverter
+        meta._ConverterMeta._bindings["entityTrigger"] = LegacyEnitityTriggerConverter
+        meta._ConverterMeta._bindings["activityTrigger"] = LegacyActivityTriggerConverter
+        meta._ConverterMeta._bindings["durableClient"] = LegacyDurableClientConverter
+    elif hasattr(pkg, 'version') and pkg.version.startswith("2."):
```
hallvictoria: nit: there's no `else` statement here. It shouldn't matter for now, but might be worth adding a log that raises a warning or falls back to legacy.

The V2 check could also be the `if` statement, and the V1 fallback becomes the `else`

azure/functions/durable_functions.py
```diff
@@ -199,3 +205,207 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
...
+        elif isinstance(obj, int):
+            return meta.Datum(type='int', value=obj)
+        elif isinstance(obj, float):
+            return meta.Datum(type='double', value=obj)
+        else:
+            raise NotImplementedError
+
+    @classmethod
+    def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
+        import azure.durable_functions as adf
```
hallvictoria: Should this call `get_durable_package` instead of the plain import?

azure/functions/durable_functions.py
```diff
@@ -199,3 +205,207 @@ def decode(cls, data: meta.Datum, *, trigger_metadata) -> typing.Any:
...
+        return adf.DurableFunctionsClient(data.value)
+
+
+def register_durable_converters():
+    """
+    Registers the appropriate Durable Functions converters based on the
+    installed Durable Functions package.
+    """
+    _logger.info("Registering Durable Functions converters based on ")
```
hallvictoria: If kept, this log should be fixed. Probably should just be `_logger.info("Registering Durable Functions converters")`?

tests/test_durable_functions.py
```diff
@@ -5,10 +5,10 @@
 import json
 
 from azure.functions.durable_functions import (
-    OrchestrationTriggerConverter,
-    EnitityTriggerConverter,
-    ActivityTriggerConverter,
-    DurableClientConverter
+    LegacyOrchestrationTriggerConverter,
```
hallvictoria, line 10: Please add some tests for the new converters as well. You can look at use `patch.dict` for mocks - `tests/decorators/test_mcp:TestMCPPackageNotInstalled` can be used as a reference

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,43 @@
...
+        _logger.info("`azure.durable_functions` package found.")
+    except ImportError:
+        _logger.info("`azure.durable_functions` package not found.")
+        return None
+
+    try:
+        if hasattr(durable_functions, 'version') and durable_functions.version.startswith("2."):
+            _logger.info("Using `azure.durable_functions` v2.x package.")
+    except Exception:
+        _logger.info("Using `azure.durable_functions` v1.x package.")
```
hallvictoria: As long as `version` is a string, neither `hasattr` or `startswith` will throw an exception. This could just be an if/else for the logs instead of a try-catch

azure/functions/decorators/durable_functions.py
andystaples: Done. Moved all non-essential logs to `debug`, and further stripped them down so only the module-selection-path logs remain (package-not-found and the detected v1.x/v2.x package). Fixed in 37714d4.

azure/functions/__init__.py
andystaples: Reverted back to `2.3.0b1` in 37714d4.

azure/functions/durable_functions.py
andystaples: Good catch. `DurableClientConverter.check_input_type_annotation` and `decode` now call `get_durable_package()` instead of a plain `import azure.durable_functions`. Fixed in 37714d4.

azure/functions/durable_functions.py
andystaples: Removed these logs. The only debug logs left in this path are the ones that reveal the module-selection decision (detected v1.x/v2.x package and which converter set gets registered). Fixed in 37714d4.

azure/functions/durable_functions.py
andystaples: Reordered so the V2 check is the `if` and the legacy path is the `else` fallback, so any package without a `2.x` version now falls back to the legacy converters. Fixed in 37714d4.

azure/functions/decorators/durable_functions.py
andystaples: Agreed — replaced the try/except with a straightforward `if/else` for the v1.x/v2.x logs. Fixed in 37714d4.

tests/test_durable_functions.py
andystaples: Added tests using the `TestMCPPackageNotInstalled` / `patch.dict` pattern: a `TestDurableFunctionsV2Converters` class covering all four v2 converters (encode/decode, type annotations, implicit output, `DurableClientConverter` with a mocked `get_durable_package`), and a `TestRegisterDurableConverters` class covering `register_durable_converters` (missing/legacy/v2 selection) and `get_durable_package` (not-installed, v2 detection, caching). Fixed in 37714d4.

azure/functions/durable_functions.py
andystaples: Fixed — the broken f-string-style message is gone. The redundant entry log was removed entirely; the remaining debug log just states which converter set is being registered. Fixed in 37714d4.

azure/functions/decorators/function_app.py
```diff
@@ -352,11 +355,12 @@ def _get_durable_blueprint(self):
         """Attempt to import the Durable Functions SDK from which DF
         decorators are implemented.
         """
-        try:
-            import azure.durable_functions as df
+        _logger.info("Getting Durable Functions blueprint.")
```
gavin-aguiar: Use debug log here,

azure/functions/decorators/function_app.py
```diff
@@ -62,6 +63,8 @@
 from azure.functions.decorators.mysql import MySqlInput, MySqlOutput, \
     MySqlTrigger
 
+_logger = logging.getLogger('azure.functions.DurableFunctions')
```
gavin-aguiar: This will set the logger for this whole file. Can we set this in the get_durable_blueprint function

azure/functions/__init__.py
```diff
@@ -48,6 +49,9 @@
 from . import connectors  # NoQA
 
 
+register_durable_converters()
```
gavin-aguiar: This will be called for every non durable function app too? Can we some only called this for durable apps?

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,39 @@
...
+    if df:
+        return df
+
+    try:
+        import azure.durable_functions as durable_functions  # noqa
+    except ImportError:
+        _logger.debug("`azure.durable_functions` package not found.")
+        return None
+
+    if hasattr(durable_functions, 'version') and durable_functions.version.startswith("2."):
```
gavin-aguiar, line 32: The current  SDK does not expose  azure.durable_functions.version. This will always log v1 package is used right?

azure/functions/durable_functions.py
```diff
@@ -5,14 +5,19 @@
 import json
 
 from azure.functions import _durable_functions
+from azure.functions.decorators.durable_functions import get_durable_package
 from . import meta
 
+import logging
```
gavin-aguiar: Add this with the other imports.

azure/functions/__init__.py
gavin-aguiar: This will trigger `import azure.durable_functions`  via  `get_durable_package()` . Since the durable SDK imports  azure.functions  at its top level, this is a re-entrant import that happens while  azure.functions  is still initializing.
Could we register lazily instead of at import time? e.g. resolve/register the converters on first binding lookup (in  _ConverterMeta.get() ) or on first use in the durable decorator path, so the durable SDK import happens only after  azure.functions  has fully finished loading.

azure/functions/decorators/durable_functions.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    global df
+    if df:
+        return df
+
+    try:
+        import azure.durable_functions as durable_functions  # noqa
+    except ImportError:
+        _logger.debug("`azure.durable_functions` package not found.")
+        return None
```
gavin-aguiar, line 30: In which case this will return none? Won't a durable app always has the SDK installed?

azure/functions/decorators/durable_functions.py
hallvictoria: AFD V2 will expose this property

azure/functions/decorators/durable_functions.py
andystaples: Good question — the `None` path is reachable and intentional in two cases:

1. **Non-durable apps (the common case).** `register_durable_converters()` runs at import of `azure.functions` for *every* app. The majority of apps don't have `azure-functions-durable` installed, so `get_durable_package()` returns `None` and registration is a deliberate no-op.

2. **A user references a durable decorator without the SDK installed.** The durable decorators (`orchestration_trigger`, `entity_trigger`, `activity_trigger` on `TriggerApi`; `durable_client_input` on `BindingApi`) are defined unconditionally in the base `azure-functions` library and inherited by every `FunctionApp`/`Blueprint`. So a user can reference `@app.orchestration_trigger` even without the SDK present. In that case `_get_durable_blueprint()` gets `None` and raises the explicit "please install `azure-functions-durable`" error, instead of surfacing a confusing `ImportError`/`AttributeError`.

So a correctly-configured durable app will always have the SDK, but `None` is the right guard for both the non-durable-app import path and the misconfigured-durable-app path.

azure/functions/__init__.py
andystaples: Great catch on the re-entrant import. Switched to lazy registration in 5ec891e: instead of calling `register_durable_converters()` at `azure.functions` import time, it's now registered as a deferred callback via `get_binding_registry().register_deferred(...)`. `_ConverterMeta.get()` runs any pending deferred registrations once, on the first binding lookup — which happens after `azure.functions` has fully finished loading. So the durable SDK import (and its `import azure.functions`) no longer occurs mid-initialization. Added tests in `test_meta.py` covering that the callback runs on first `get()` and only once.

azure/functions/decorators/durable_functions.py
andystaples: Correct — with the current published SDK there's no `version` attribute, so it takes the `else` branch and logs v1.x, which is the intended behavior today. The `version`-based check is forward-compat: once the v2.x (Durable Task) SDK ships exposing `version = "2.x"`, the same code will detect and log v2.x and register the new converters. It's a debug-only log, so no functional impact either way.

azure/functions/durable_functions.py
andystaples: Done — `import logging` is now grouped with the other stdlib imports (`json`, `typing`) at the top of the file in 5ec891e.

azure/functions/decorators/function_app.py
andystaples: Good point — the module-level logger was only used in `_get_durable_blueprint`. Moved it inside that function in 5ec891e.

azure/functions/__init__.py
andystaples: Addressed together with your re-entrant-import comment below — registration is now deferred to the first binding lookup instead of running at import time for every app. See 5ec891e.

azure/functions/decorators/function_app.py
andystaples: Done — changed to `logger.debug(...)` in 5ec891e.

andystaples:
Thanks @gavin-aguiar for the review. 
The agent might have gone a bit overboard with the changes to meta.py and deferred registration - let me know and I can revert and go for a more focused fix

PR #354, opened 2026-07-24
feat: add register_converter API

hallvictoria:
The `azure-functions-durable` v2.x SDK uses the durabletask gRPC programming model, which requires orchestration and entity trigger outputs to be encoded as string datums rather than json. Previously there was no public way to override a registered converter.

This change introduces a minimal, stable public API that allows `azure-functions-durable` (and only that package) to replace durable-related converters without touching any private internals of `azure-functions`.

Only 4 binding names are permitted to be overridden: `orchestrationTrigger`, `entityTrigger`, `activityTrigger`, `durableClient`.

This does not change:
1. No changes to the cold-start path for non-durable function apps
2. No changes to existing converter registration behavior
3. Converter overrides are the responsibility of `azure-functions-durable` itself

### DIFF
--- a/azure/functions/__init__.py
+++ b/azure/functions/__init__.py
@@ -23,7 +23,7 @@ from ._http_wsgi import WsgiMiddleware
 from ._http_asgi import AsgiMiddleware
 from .kafka import KafkaEvent, KafkaConverter, KafkaTriggerConverter
 from .mcp import MCPToolContext, PromptInvocationContext
-from .meta import get_binding_registry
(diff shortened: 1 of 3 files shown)


andystaples:
FWIW, this works really well for the Durable scenario and removes the ask for PRs like
https://github.com/Azure/azure-functions-python-library/pull/353
https://github.com/Azure/azure-functions-python-library/pull/309

gavin-aguiar (approved):

